PR #37, opened 2025-05-18
feat: add base URL configuration for providers

missuo:
- Allow users to configure a custom `Base URL`.

For #33

### DIFF
--- a/src/entrypoints/options/providers/base-url-config.tsx
+++ b/src/entrypoints/options/providers/base-url-config.tsx
@@ -1,0 +1,62 @@
+import type { APIProviderNames } from '@/types/config/provider'
+import { useAtom } from 'jotai'
+import { useCallback } from 'react'
+import ProviderIcon from '@/components/provider-icon'
+import { Input } from '@/components/ui/input'
+import { configFields } from '@/utils/atoms/config'
+import { API_PROVIDER_ITEMS } from '@/utils/constants/config'
+
+export default function BaseURLConfig() {
+  const [providersConfig, setProvidersConfig] = useAtom(configFields.providersConfig)
+
+  const handleBaseUrlChange = useCallback((provider: APIProviderNames, newBaseUrl: string) => {
+    setProvidersConfig({
+      [provider]: {
+        ...providersConfig[provider],
+        baseURL: newBaseUrl,
+      },
+    })
+  }, [setProvidersConfig, providersConfig])
+
+  // get placeholder for each provider
+  const getPlaceholder = (provider: string) => {
+    switch (provider) {
+      case 'openrouter':
+        return 'https://openrouter.ai/api/v1'
+      case 'openai':
+        return 'https://api.openai.com/v1'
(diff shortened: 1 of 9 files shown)


src/entrypoints/options/providers/base-url-config.tsx
```diff
@@ -0,0 +1,62 @@
...
+    setProvidersConfig({
+      [provider]: {
+        ...providersConfig[provider],
+        baseURL: newBaseUrl,
+      },
+    })
+  }, [setProvidersConfig, providersConfig])
+
+  // get placeholder for each provider
+  const getPlaceholder = (provider: string) => {
```
mengxi-ream: is there any other way to not code this in the component level? like move the logic to constants folder, or don't need switch is even better

src/utils/provider.ts
```diff
@@ -14,9 +14,11 @@ export async function getProviderRegistry() {
 
   return createProviderRegistry({
     openai: createOpenAI({
+      baseURL: config?.providersConfig?.openai.baseURL ?? 'https://api.openai.com/v1',
       apiKey: config?.providersConfig?.openai.apiKey,
     }),
     deepseek: createDeepSeek({
+      baseURL: config?.providersConfig?.deepseek.baseURL ?? 'https://api.deepseek.com/v1',
```
mengxi-ream: seems duplicated code as the switch logic in the component file. if we code this way, then need to change the code both side when change something in the future.

mengxi-ream (changes requested):
Hi, thanks for the PR!!! Really appreciate it.

Additionally to the comment. you may need to:
1. add the config schema migration function and test it (check example at folder: utils/config)
2. add changeset record

mengxi-ream:
Another point I would like to discuss is the following:

If we initially set the baseURL to undefined, but we don't want users to see an empty input in the frontend UI, as it can be ambiguous compared to setting the baseURL as an empty string.

I noticed that you used a default value in the UI. However, do you think it would be better to define the baseURL as a string and initialize it with the official baseURL in DEFAULT_PROVIDER_CONFIG? This way, we wouldn't need additional logic in the UI component or registry to handle the default value fallback.

missuo:
> Another point I would like to discuss is the following:
> 
> If we initially set the baseURL to undefined, but we don't want users to see an empty input in the frontend UI, as it can be ambiguous compared to setting the baseURL as an empty string.
> 
> I noticed that you used a default value in the UI. However, do you think it would be better to define the baseURL as a string and initialize it with the official baseURL in DEFAULT_PROVIDER_CONFIG? This way, we wouldn't need additional logic in the UI component or registry to handle the default value fallback.

I agree with you. We don't want users to see an empty input, nor do we want them to have to enter a baseURL. Therefore, it is essential to initialize a default baseURL as a value; this is the best choice.

Give me some time to optimize the current code.

mengxi-ream:
lgtm, thanks for the PR :)

mengxi-ream (approved):

Issue #52, opened 2025-05-20
[FEATURE] Integrate Ollama


zmrlft:
**Is your feature request related to a problem? Please describe.**
Whenever I am worried about data leakage, I always hope to deploy a large model locally to help me with translation

**Describe the solution you'd like**
I hope to integrate Ollama. In this way, I can conveniently use various large models deployed locally for translation without worrying about data leakage

**Describe alternatives you've considered**
no

**Additional context**
I hope it can be assigned to me. I know this might be difficult to complete, but I want to give it a try. Besides, I have an rtx2060 12GB graphics card and can deploy many large models for testing

PR #61, opened 2025-05-23
feat: the feature integrate ollama

zmrlft:
## Type of Changes
- [x] ✨ New feature (feat)
- [x] 🧠 AI model related (ai)

## Description
This PR integrates the Ollama local large model service into the read-frog project, enabling offline translation capabilities using locally deployed large models.        Users can now leverage models like `llama2` and `gemma3` deployed via Ollama for translation tasks without relying on third-party API services, enhancing data privacy and offline usability.

## Related Issue
Closes
https://github.com/mengxi-ream/read-frog/issues/52

## How Has This Been Tested?
- [x] Manual testing: Verified that Ollama models can be pulled/run locally and correctly return translation results through the read-frog interface.

## Checklist
- [x] I have tested these changes locally (Ollama models deployed and translated successfully in Windowsenvironments).
- [x] I have updated the documentation: I added Ollama-related documents in the "docs" directory.
- [x] My code follows the project's TypeScript style guide and maintains compatibility with existing provider interfaces.

## Additional Information
- Ollama model support is added to translation functions (note: read function may have context window limitations. Most local models do not support the reading function. So the reading function still needs to be improved by opening a new feat).
- Default recommended model: `gemma3:1b` (815MB, suitable for low-resource environments).
- The cross-domain issue has currently been resolved by configuring the ollama server environment variable.  However, we can also enable the read frog plugin to support cross-domain, which requires opening a new feat


.wxt/i18n/structure.d.ts
```diff
@@ -0,0 +1,65 @@
+export type GeneratedI18nStructure = {
```
mengxi-ream: .wxt should be git ignored

src/utils/config/__tests__/example/v004.ts
```diff
@@ -17,6 +17,9 @@ export const configExample: Config = {
     openrouter: {
       apiKey: undefined,
     },
+    ollama: {
```
mengxi-ream: You shouldn't change either `v004.ts` or `v005.ts`. because your config schema version should be 6

src/utils/host/translate-text.ts
```diff
@@ -45,7 +45,16 @@ export async function translateText(sourceText: string) {
         cleanSourceText,
       ),
     })
-    translatedText = text
+    // Some deep thinking models, such as deepseek, return the thinking process. Therefore,
+    // the thinking process in the <think></think> tag needs to be filtered out and only the result is returned
+    if (text.includes('<think>')) {
```
mengxi-ream: how long will a think model take to translate a sentence?

src/utils/provider.ts
```diff
@@ -34,6 +38,13 @@ export async function getTranslateModel(provider: keyof typeof translateProvider
   if (provider === 'openrouter') {
     return openrouter.languageModel(model)
   }
+  if (provider === 'ollama') {
```
mengxi-ream: no, ollama !== openrouter, shouldn't use createOpenRouter

mengxi-ream (changes requested):
In additional to all the comments. You also need to 
1. write the test for config schema migration from 5 to 6
2. record a video show that all the Ollama models you put in the file `types/config/provider.ts` can handle the translation and reading tasks.
3. add a changeset record

src/utils/provider.ts
zmrlft: Hello. Thank you for your comment. openrouter is also compatible with ollama and can proxy call the locally running Ollama. I think there are several benefits to writing it this way:
All providers (openai, deepseek, ollama) use the unified interface and api of.languageModel(model)
2. There is no need to introduce additional SDKS (such as ollama's ai development kit), avoiding the complication of dependency management
3. It simplifies the code structure and maintenance costs

src/utils/host/translate-text.ts
zmrlft: My graphics card is the 12GB version of rtx2060, with 16GB of memory. Under the Windows 10 environment, I deployed the deepseek r1:8b large model to translate a sentence. It will output the reasoning and thinking process and finally provide the translation result. The time takes about 3 seconds. Depending on the length of the sentence and the system load, the time will vary accordingly

src/utils/provider.ts
mengxi-ream: how about https://ai-sdk.dev/providers/community-providers/ollama? This is the official way. OpenRouter's provider has some compatible issue with ai sdk registry, that's why I'm reluctant to use

zmrlft: There are also compatibility issues between ollama and the ai sdk registry. If I write it like this, the local ollama will report an error 404, and it cannot understand this request image 
image

mengxi-ream: it might be the bug from their side. What if you use createOllma outside of registry like openrouter

mengxi-ream: > Hello. Thank you for your comment. openrouter is also compatible with ollama and can proxy call the locally running Ollama. I think there are several benefits to writing it this way: All providers (openai, deepseek, ollama) use the unified interface and api of.languageModel(model) 2. There is no need to introduce additional SDKS (such as ollama's ai development kit), avoiding the complication of dependency management 3. It simplifies the code structure and maintenance costs

We shouldn't use createOpenrouter because it will change in the future and can't guarantee it will always be compatible with Ollama, ... and, it's ugly.

zmrlft: For example, write like this: 
image 
It will also be 404.
image

mengxi-ream: maybe just return the model like this 

```
createOllama({
  // custom settings
});
```

no `.languageModel`

or you can post an issue there to ask for help https://github.com/sgomez/ollama-ai-provider/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen

zmrlft: > maybe just return the model like this
> 
> ```
> createOllama({
>   // custom settings
> });
> ```
> 
> no `.languageModel`
> 
> or you can post an issue there to ask for help https://github.com/sgomez/ollama-ai-provider/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen

the 404 error occurred as well

zmrlft: > > Hello. Thank you for your comment. openrouter is also compatible with ollama and can proxy call the locally running Ollama. I think there are several benefits to writing it this way: All providers (openai, deepseek, ollama) use the unified interface and api of.languageModel(model) 2. There is no need to introduce additional SDKS (such as ollama's ai development kit), avoiding the complication of dependency management 3. It simplifies the code structure and maintenance costs
> 
> We shouldn't use createOpenrouter because it will change in the future and can't guarantee it will always be compatible with Ollama, ... and, it's ugly.

Yes you are considering, but now the problem is really difficult, the current solution is based on the following link to solve: https://ollama.ac.cn/blog/openai-compatibility
Since the ollama server currently supports calling with openai's interface format, if the ollama official continues to support this OpenAI-compatible function, it is feasible for us to call ollama with openai
image

mengxi-ream (changes requested):
looks good to me. may need to do two things:
1. add changeset by `pnpm changeset`
2. find a larger ollama icon like 128*128 or 256*256 (don't be too large)

src/utils/host/translate-text.ts
```diff
@@ -45,9 +45,13 @@ export async function translateText(sourceText: string) {
         cleanSourceText,
       ),
     })
-    translatedText = text
+    // Some deep thinking models, such as deepseek, return the thinking process. Therefore,
+    // the thinking process in the <think></think> tag needs to be filtered out and only the result is returned
+    const [, extracted = text] = text.match(/<\/think>([\s\S]*)/) || []
+    const cleaned = extracted.trim().replace(/\s+/g, '')
+    return cleanSourceText === cleaned ? '' : cleaned
```
mengxi-ream: don't return here, it's redundant code, you should:
```
translatedText = extracted.trim().replace(/\s+/g, '')
```
Then just it, the return logic is in the line 56

mengxi-ream (changes requested):

src/utils/host/translate-text.ts
zmrlft: > don't return here, it's redundant code, you should:
> 
> ```
> translatedText = extracted.trim().replace(/\s+/g, '')
> ```
> 
> Then just it, the return logic is in the line 56

ok

mengxi-ream (approved):

PR #258, opened 2025-08-10
fix: Firefox Radix UI Select compatibility with preview registration

### LINKED ISSUE #57: Firefox Compatibility
To make read frog compatible to firefox

It is required that almost all functions work properly on Firefox, with a few exceptions.

> \[!IMPORTANT]
>
> 💸 Developer Incentive: Anyone who implements this (in line with current code specifications) will receive a 🧧 $100 (USD) cash incentive.
> 💸 开发者激励：任何按照当前代码规范实现的人将获得 🧧 $100 美金现金激励。（可折算人民币）

Haleclipse:
## Summary

- Fix Firefox Radix UI Select compatibility issues in popup showing localized text and icons
- Create browser-adaptive-select component with Firefox fallback and Chrome Radix passthrough  
- Implement ItemText preview registration system to display correct SelectItem content instead of raw values
- Resolve popup state persistence issues through hidden content mounting

### Core Implementation
- Add `apps/extension/src/components/browser-adaptive-select.tsx`
  - Browser-adaptive Select component detecting Firefox and providing custom implementation
  - ItemText preview registration system capturing SelectItem children content
  - Hidden content mounting pattern maintaining preview state across popup sessions

### Code Quality  
- Clean up Chinese comments in Firefox compatibility fix files
- Unify code style and comment language

### Preview Registration System
Store SelectItem React node content via Map:
```tsx
const setItemPreview = useCallback((v: V, node: React.ReactNode) => {
  previewsRef.current.set(v, node)
}, [])
```

### Hidden Content Mounting
Maintain children rendering when Select is closed to preserve preview registration:
```tsx
if (\!open) {
  return (
    <div style={{ display: 'none' }} aria-hidden>
      <div className="p-1">{children}</div>
    </div>
  )
}
```

## Test Plan

- [x] Firefox MV3 build passes
- [x] SelectValue displays correct localized text and icons
- [x] Popup reopen maintains correct state
- [x] Chrome environment continues using native Radix UI

## Related

Closes #57


mengxi-ream:
Hi, thank you for the pull request. I'm not quite sure I understand why we need a new component. Could you please set up a meeting for me to explain the PR if you are avaiblable? https://calendar.notion.so/meet/mengxi-ream/tk54c4omv

The main concern is that I wanna make sure this pattern of code is maintainable for the future development as well.

Haleclipse:
> Hi, thank you for the pull request. I'm not quite sure I understand why we need a new component. Could you please set up a meeting for me to explain the PR if you are avaiblable? https://calendar.notion.so/meet/mengxi-ream/tk54c4omv
> 
> The main concern is that I wanna make sure this pattern of code is maintainable for the future development as well.

like them

https://lightrun.com/answers/cypress-io-cypress-radix-ui-select-component-breaks-in-firefox-tests
https://github.com/radix-ui/primitives/issues/2754

Radix UI's Select does not work normally in the Popup.

So I re-simulated its behavior and wrote a similar styled Select component, and kept using the original Radix UI Select under Chrome.

image

You can compare and test it.

apps/extension/src/entrypoints/background/index.ts
```diff
@@ -55,6 +55,45 @@ export default defineBackground(() => {
     }
   })
 
+  // Firefox-specific fetch proxy for content script CSP restrictions
+  onMessage('proxyFetch', async (message) => {
```
mengxi-ream: could you extract this logic to another single file?

apps/extension/src/entrypoints/popup/components/language-level-selector.tsx
```diff
@@ -8,7 +8,7 @@ import {
   SelectItem,
   SelectTrigger,
   SelectValue,
-} from '@/components/ui/select'
```
mengxi-ream: could you create a component wrapper that if on the browser that is not Firefox, use the Radix original component, if on the browser that is Firefox, use this adaptive component. so this ensure component on other browser won't be possibly broken

apps/extension/src/utils/fixes/firefox-icons.ts
```diff
@@ -0,0 +1,66 @@
...
+export function applyFirefoxIconsFix() {
+  // Check if running in Firefox environment - apply fixes for non-Chrome/Edge browsers
+  if (import.meta.env.BROWSER !== 'firefox') {
+    return
+  }
+
+  // Preload icons actually used in floating-button to resolve Firefox CSP restrictions
+
+  // Icons used in floating-read-button.tsx
+  addIcon('tabler:book', {
```
mengxi-ream: Is this really necessary for Firefox? If so, it might be better to switch the icon library to "Lucide" or "Tabler" directly? as these libraries are compatible with all browsers and would eliminate the need for any special logic for Firefox.

apps/extension/src/utils/fixes/firefox-fetch.ts
```diff
@@ -0,0 +1,149 @@
+import { sendMessage } from '@/utils/message'
+
+/**
+ * Firefox network request fix
+ * Content scripts cannot make direct fetch requests due to Firefox CSP restrictions
```
mengxi-ream: Does this restriction only exist in Firefox?

apps/extension/src/entrypoints/side.content/index.tsx
```diff
@@ -33,6 +35,11 @@ export default defineContentScript({
   cssInjectionMode: 'ui',
   async main(ctx) {
     await loadGlobalConfig()
+
+    // Apply Firefox compatibility fixes
+    applyFirefoxIconsFix()
+    applyFirefoxFetchFix()
```
mengxi-ream: if these fixes are necessary, should them applied to all content scripts?

apps/extension/src/utils/styles.ts
```diff
@@ -17,8 +17,33 @@ export function addStyleToShadow(shadow: ShadowRoot) {
...
+  let mirrorSheet: CSSStyleSheet | null = null
+  let mirrorStyleElement: HTMLStyleElement | null = null
+
+  if (supportsAdoptedStyleSheets) {
+    // Modern browsers: use adoptedStyleSheets
+    mirrorSheet = new CSSStyleSheet()
+    shadowRoot.adoptedStyleSheets = [...shadowRoot.adoptedStyleSheets, mirrorSheet]
+  }
+  else {
+    // Compatibility mode: use <style> element
```
mengxi-ream, line 34: I think Firefox support adoptedStyleSheets, is there any reason for this?

mengxi-ream:
Thank you very much for the pull request! You seem to have a lot of knowledge about Firefox. really appreciate your effort to make this happen. I left some comments~

apps/extension/src/entrypoints/background/index.ts
Haleclipse: Oh, it really should be merged into the same function as `backgroundFetch`.

apps/extension/src/entrypoints/popup/components/language-level-selector.tsx
Haleclipse: Here should be considered my mistake.

My rollback logic directly points to the more primitive react-select referenced by Radix UI

Initially, it was to simulate a custom Select

Actually, it should roll back to using `components/ui/select.tsx`

mengxi-ream: okok, I prefer not to affect the code on Chrome and Edge side, so make sure the new change won't break the existing features.

apps/extension/src/utils/fixes/firefox-icons.ts
Haleclipse: It is very necessary to preload the SVG code of the icons used due to incompatibility between the loading mechanism of this icon and Shadow DOM. Otherwise, they will not display normally. You can try to comment out the call to `applyFirefoxIconsFix` to observe the problem. On Firefox's Shadow DOM, only an empty content placeholder `<span></span>` can be loaded.

mengxi-ream: Thanks for the insight! If that's the case, shouldn't we just use `lucide` and `tabler` or `remix-icon` libraries directly and remove `iconify` from our repo?

apps/extension/src/utils/styles.ts
Haleclipse: Explanation for the Array Spread Workaround

  The Root Cause - MV2 Legacy Issue:

  The array spread workaround [...shadowRoot.adoptedStyleSheets, mirrorSheet] was originally implemented because:

  1. WXT Default Behavior: WXT initially defaulted to Manifest V2 builds
  2. MV2 Browser Bug: In Manifest V2 context, shadowRoot.adoptedStyleSheets.push() threw TypeError: adoptedStyleSheets.push is not
   a function
  3. Compatibility Fix: We implemented the array spread as a defensive measure to handle this MV2-specific issue

  The Migration Context:

  - Original Development: Started with MV2 where the bug existed
  - Forced Migration: Due to browser deprecation timelines, we were forced to migrate to MV3
  - Forgotten Workaround: During the MV3 migration, this compatibility fix was overlooked and left in place

  Current Reality:

  - MV3 Support Confirmed: You've verified that MV3 properly supports adoptedStyleSheets.push()
  - Legacy Code: The array spread is now unnecessary technical debt from the MV2 era
  - Clean Solution: We can safely revert to the simpler push() method

  Summary:

  This was a historical compatibility patch for MV2 that became obsolete when we migrated to MV3, but was inadvertently carried
  forward. The workaround solved a real problem in MV2 but is no longer needed in the modern MV3 environment.

apps/extension/src/utils/fixes/firefox-icons.ts
Haleclipse: If you are willing, you can do it this way

Using only specific range of icons can also make the package smaller

Of course, I am just making PR patches based on two principles

1. Not intruding too much on the original logic, making it easy to remove when the browser truly supports it

2. Not directly modifying the component code generated by the component library used, but providing a copy and integrating the modifications

_So if these points are not considered when making repairs, in fact, it only takes less time and effort._

apps/extension/src/utils/styles.ts
mengxi-ream: Haven't we already worked on the MV3 environment? Or are you saying that the Firefox extension needs MV2?

Based on your response, it seems that code is not necessary for MV3, which I believe we are currently adopting.

Please correct me if I'm wrong?

Parsifa1: really weird:
https://github.com/wxt-dev/wxt/issues/1195#issuecomment-2536633336
seems that there are some error about `adoptedStyleSheets` in firefox although it do support it.

apps/extension/wxt.config.ts
```diff
@@ -6,7 +6,36 @@ export default defineConfig({
   imports: false,
   modules: ['@wxt-dev/module-react', '@wxt-dev/i18n/module'],
   vite: () => ({
-    plugins: [],
+    plugins: [
```
Parsifa1: no need to do this, just using `zod/v4` and set `z.config({jitless: true})`

ananaBMaster: Good catch! Our team has sent you an email at your address, i@cloudti.de. Let's connect! @Parsifa1

Haleclipse: thanks

ananaBMaster: zod has been upgraded here: https://github.com/mengxi-ream/read-frog/pull/279

mengxi-ream:
After removing the svg icon pattern and import these directly from Firefox, then please request me to review again at Github.

Haleclipse:
check

wenwen1145141919810:
这个pr啥时候能合啊，现在只能先用kiss translator了

apps/extension/package.json
```diff
@@ -6,10 +6,10 @@
   "description": "Read Frog browser extension for language learning",
   "scripts": {
     "build": "wxt build",
-    "build:firefox": "wxt build -b firefox",
+    "build:firefox": "wxt build -b firefox --mv3",
     "dev": "wxt",
     "dev:edge": "wxt -b edge",
-    "dev:firefox": "wxt -b firefox",
+    "dev:firefox": "wxt -b firefox --mv3",
```
mengxi-ream, line 12: can't run dev mode in mv3 actually

apps/extension/src/utils/zod-csp.ts
```diff
@@ -0,0 +1,8 @@
+// Zod CSP compatibility configuration for Firefox extensions
+// This must be imported before any Zod schemas are created
+import { z } from 'zod'
```
mengxi-ream, line 3: It’s not ideal if we need to import { z } from this custom file. Is there a way to import from 'zod' while allowing only Firefox with the jitless setting?

apps/extension/src/components/browser-adaptive-select.tsx
```diff
@@ -0,0 +1,630 @@
...
+  SelectItem as UISelectItem,
+  SelectLabel as UISelectLabel,
+  SelectTrigger as UISelectTrigger,
+  SelectValue as UISelectValue,
+} from '@repo/ui/components/select'
+import * as React from 'react'
+import { useCallback, useEffect, useId, useLayoutEffect, useMemo, useRef, useState } from 'react'
+import { createPortal } from 'react-dom'
+
+const isFirefox = typeof navigator !== 'undefined' && /firefox/i.test(navigator.userAgent)
```
mengxi-ream, line 14: use wxt's way to determine if it's firefox

apps/extension/src/components/browser-adaptive-select.tsx
```diff
@@ -0,0 +1,630 @@
...
+  disabled?: boolean
+  placeholder?: React.ReactNode
+  setItemPreview: (v: V, node: React.ReactNode) => void
+  getItemPreview: (v?: V) => React.ReactNode | undefined
+}
+
+const Ctx = React.createContext<ContextType | null>(null)
+function useCtx() {
+  const c = React.use(Ctx)
+  if (!c)
```
mengxi-ream, line 40: The variable names `v`, `b`, `cv`, `uv`, and `c` lack specificity and clarity for other developers. We do not adhere to this naming convention. Open source projects involve collaboration, not just individual contributions, so we need a more effective way to ensure clarity.

apps/extension/src/components/browser-adaptive-select.tsx
```diff
@@ -0,0 +1,630 @@
...
+
+const Fallback = { Root, Trigger, Value, Content, Viewport, Item, ItemText, Portal, Icon, Group, Label } as const
+
+// UI components adapter for non-Firefox browsers
+const UIComponents = {
+  Root: UISelect,
+  Trigger: UISelectTrigger,
+  Value: UISelectValue,
+  Content: UISelectContent,
+  Viewport: ({ children }: React.PropsWithChildren) => <>{children}</>, // ui/select handles viewport internally
```
mengxi-ream, line 612: Why is `Viewport` defined here but not exported? Is this legacy code from an AI?

apps/extension/src/components/browser-adaptive-select.tsx
```diff
@@ -0,0 +1,630 @@
...
+      <div
+        className="p-1 select-scroll-container"
+        style={{
+          flex: 1,
+          overflowY: 'auto',
+          scrollbarWidth: 'none', /* Firefox */
+          msOverflowStyle: 'none', /* IE and Edge */
+        } as React.CSSProperties}
+      >
+        <style>
```
mengxi-ream, line 487: write the <style> directly inside component?

apps/extension/src/components/browser-adaptive-select.tsx
```diff
@@ -0,0 +1,630 @@
...
+    >
+      <ScrollUpButton />
+      <div
+        className="p-1 select-scroll-container"
+        style={{
+          flex: 1,
+          overflowY: 'auto',
+          scrollbarWidth: 'none', /* Firefox */
+          msOverflowStyle: 'none', /* IE and Edge */
+        } as React.CSSProperties}
```
mengxi-ream, line 485: is there particular reason not using tailwind here?

mengxi-ream: avoid `as` assertion unless really necessary

apps/extension/src/components/browser-adaptive-select.tsx
```diff
@@ -0,0 +1,630 @@
...
+
+  useEffect(() => {
+    setItemPreview(value, <>{children}</>)
+  }, [value, children, setItemPreview])
+
+  return <span>{children}</span>
+}
+const Icon: React.FC<React.PropsWithChildren<Record<string, never>>> = ({ children }) => <span aria-hidden>{children ?? '▾'}</span>
+
+function Group({ children, className, ...props }: any) {
```
mengxi-ream, line 583: avoid `any` type

mengxi-ream (changes requested):
browser adapative selection is to solve the compatibility issue, but why design browser adapative selection in this particular way. how to maintain the browser adapative selection in the future if the radix change some api. are you sure all radix api are compatible, please explain and you can schedule a meeting with me if you feel necessary.  https://calendar.notion.so/meet/mengxi-ream/tk54c4omv

YandLiu:
@Haleclipse 可以上传个Firefox版让我们先尝个鲜吗？

Issue #170, opened 2025-06-24
[BUG] Read Drawer Auto-Opens and Cannot Be Closed


HoikanChan:
# Bug Report

**Describe the bug**
When entering a specific website, the translation plugin automatically opens a drawer/sidebar that cannot be closed. The close button is unresponsive and the drawer remains open persistently.

**To Reproduce**
1. Navigate to the specific website (URL appears to contain translation-related content. like https://langchain-ai.github.io/langgraph/tutorials/workflows/#orchestrator-worker)
2. The translation plugin drawer automatically opens on the right side showing "Workflows & agents" with a "Read" button
3. Attempt to close the drawer by clicking the close button (X)
4. The drawer remains open and cannot be dismissed

**Expected behavior**
The translation plugin drawer should either:
- Not automatically open when visiting the website, OR
- Be able to be closed when the user clicks the close button

**Device (please complete the following information):**
- Device:  Desktop PC,
- OS: Windows 11
- Browser:  Chrome  137
- App Version: 0.10.9

![Image](https://github.com/user-attachments/assets/6d8afa7b-1022-47eb-bdb0-827b12598df9)

PR #420, opened 2025-09-02
fix: reinsert when an internal node is removed

### LINKED ISSUE #410: [BUG] 网站通过 url 锚点改变内容（即单页应用）点击改变锚点的链接时，侧边栏会弹出来
### Before you report

点击导航链接时，侧边栏弹出，而且没法点击关闭按钮关掉侧边栏

### To Reproduce

如： https://docs.basedpyright.com/v1.29.4/benefits-over-pyright/new-diagnostic-rules/
点击左侧导航链接，侧边栏就会弹出

Browser Extension

1.0.0 ~ 1.5.4

```bash
- Edge ：  139.0.3405.119
- Windows 11
```

### Additional context

_No response_

taiiiyang:
## Type of Changes

- [ ] ✨ New feature (feat)
- [x] 🐛 Bug fix (fix)
- [ ] ✅ Test related (test)
- [ ] 🔄 CI/CD related (ci)
- [ ] 🧠 AI model related (ai)

## Description

 reinsert when an internal node is removed in side.content

## Related Issue

Closes #410 #170 

## How Has This Been Tested?

- [ ] Added unit tests

## Checklist

## Additional Information

    

---

## Summary by cubic
Prevents style loss when internal extension style nodes are removed by reinserting them, fixing side content auto-expand styling issues.

- **Bug Fixes**
  - Detects extension-owned style nodes (wxt shadow styles, _goober, sonner toaster) with a new isInnerDom check.
  - On DOM changes, reattaches any removed internal nodes to document.head within mirrorDynamicStyles.

### DIFF
--- a/apps/extension/src/utils/styles.ts
+++ b/apps/extension/src/utils/styles.ts
@@ -27,7 +27,7 @@ function isInnerDom(node: Node) {
     return true
   }
 
-  if (node.textContent?.includes('[data-sonner-toaster]')) {
+  if (node instanceof HTMLStyleElement && node.textContent.includes('[data-sonner-toaster]')) {
     return true
   }
 
@@ -77,6 +77,7 @@ export function mirrorDynamicStyles(selector: string, shadowRoot: ShadowRoot, co
   // Observe the head for added style elements
   const headObserver = new MutationObserver((mutations) => {
     mutations.forEach((mutation) => {
+      console.log(mutation.addedNodes)
       mutation.addedNodes.forEach((node) => {
         if (node instanceof HTMLStyleElement && node.matches(selector)) {
           // Only check content if contentMatch is provided
--- a/apps/extension/src/utils/styles.ts
+++ b/apps/extension/src/utils/styles.ts
@@ -77,7 +77,6 @@ export function mirrorDynamicStyles(selector: string, shadowRoot: ShadowRoot, co
   // Observe the head for added style elements
   const headObserver = new MutationObserver((mutations) => {
     mutations.forEach((mutation) => {
-      console.log(mutation.addedNodes)
       mutation.addedNodes.forEach((node) => {
         if (node instanceof HTMLStyleElement && node.matches(selector)) {
           // Only check content if contentMatch is provided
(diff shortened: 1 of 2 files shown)


apps/extension/src/utils/styles.ts
```diff
@@ -15,6 +15,25 @@ export function addStyleToShadow(shadow: ShadowRoot) {
...
+
+  if (node instanceof HTMLStyleElement && node.attributes.getNamedItem('wxt-shadow-root-document-styles')) {
+    return true
+  }
+
+  if (node instanceof HTMLStyleElement && node.id === '_goober') {
+    return true
+  }
+
+  if (node.textContent?.includes('[data-sonner-toaster]')) {
```
mengxi-ream: this function looks like hard coding the conditions. what if in the future we add other style elements? or we delete some styles?

We may need a systematically way. to let this function can dynamically detect all css we insert. i.e., when we insert a new style into dom in the future, we don't need to change any code in this function, and that will be automatically watched by `mutation.removedNodes`

One idea I have is to add the special `id` like `read-frog-insert-style` to every css we insert and this function can only detect that `id`, if the style has that `id` then return `true`

But there might be a better idea you can come up with. the above one just give you an example that what we want.

mengxi-ream (changes requested):
have we tested this on another similar issue as well? https://github.com/mengxi-ream/read-frog/issues/170

taiiiyang:
> have we tested this on another similar issue as well? #170

yes, no problem

apps/extension/src/utils/styles.ts
taiiiyang: I have also considered this issue, but these styles are all injected by wxt and sonner. I don't know where to uniformly intercept them.

mengxi-ream: If you don't feel there is a very good way to do. Maybe just leave it as this. But I'm not sure why we use the name "innerXXX" as the function name

taiiiyang: > If you don't feel there is a very good way to do. Maybe just leave it as this. But I'm not sure why we use the name "innerXXX" as the function name

Used to determine if it is a DOM injected internally by read frog. Can you provide a better example?

taiiiyang: > If you don't feel there is a very good way to do. Maybe just leave it as this. But I'm not sure why we use the name "innerXXX" as the function name

Can I use `isInternalStyleElement` ?

Issue #639, opened 2025-10-23
[FEATURE] support manually switch theme (light or dark) in extension


ananaBMaster:
### Before you begin

- [x] I have searched existing issues to avoid duplicates
- [ ] I am willing to contribute a PR for this feature

### Is your feature request related to a problem? Please describe.

support manually switch theme (light or dark) in extension

### Describe the solution you'd like

switch in options page.

### Which area(s) would this feature affect? (Select all that apply)

Browser Extension

zihenzzz:
Hello, is anyone fulfilling this requirement? Can I try it out?

mengxi-ream:
sure, you need to explore theme provider and 

```
export function isDarkMode() {
  return (
    // TODO: change this to storage API for browser extension
    localStorage.theme === 'dark'
    || (!('theme' in localStorage)
      && typeof window !== 'undefined'
      && window.matchMedia
      && window.matchMedia('(prefers-color-scheme: dark)').matches)
  )
}

```

we should have "light" "dark" and "system" modes

PR #725, opened 2025-11-22
feat: support manually switch theme (light or dark) in extension

zihenzzz:
## Type of Changes

- [x] ✨ New feature (feat)

## Description

Added manual theme switching functionality to the extension, supporting three modes:

## Related Issue

Closes #https://github.com/mengxi-ream/read-frog/issues/639

## How Has This Been Tested?

## Screenshots

594f65f757a607fd308d06b3fbf1515c
8cb474ebe1af88ac0e7c073f5b7edcc5

## Checklist

---
## Summary by cubic
Added manual theme switching with light, dark, and system modes. The choice is saved and applied across the popup, options, and content UIs; system mode follows OS changes.

- **New Features**
  - Added a ThemeManager that persists mode to storage and broadcasts updates via a themechange event.
  - Updated ThemeProvider to expose themeMode and setThemeMode, resolve the effective theme, and listen to system changes only in system mode.
  - Added a ThemeConfig section in Options → General with a Light/Dark/System selector.
  - Wrapped content scripts with Jotai and ThemeProvider so UI updates immediately across views.

<sup>Written for commit 564a359eea9cdcc2283057ec20fbf4d14a19b9ee. Summary will update automatically on new commits.</sup>

### DIFF
--- a/src/utils/theme-manager.ts
+++ b/src/utils/theme-manager.ts
@@ -1,0 +1,59 @@
+import { storage } from '#imports'
+import { THEME_MODE_KEY } from './constants/storage-keys'
+
+export type ThemeMode = 'light' | 'dark' | 'system'
+
+let cachedThemeMode: ThemeMode = 'system'
+let isInitialized = false
+
+function getSystemTheme(): 'light' | 'dark' {
+  if (typeof window !== 'undefined' && window.matchMedia) {
+    return window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light'
+  }
+  return 'light'
+}
+
+export async function initThemeManager(): Promise<void> {
+  if (isInitialized)
+    return
+
+  try {
+    const stored = await storage.getItem<ThemeMode>(`local:${THEME_MODE_KEY}`)
+    if (stored && ['light', 'dark', 'system'].includes(stored)) {
+      cachedThemeMode = stored
+    }
+  }
+  catch (error) {
+    console.error('Failed to load theme from storage:', error)
+  }
+
+  isInitialized = true
+
+  storage.watch<ThemeMode>(`local:${THEME_MODE_KEY}`, (newValue) => {
+    if (newValue && ['light', 'dark', 'system'].includes(newValue)) {
+      cachedThemeMode = newValue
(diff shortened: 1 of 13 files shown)


src/utils/theme-manager.ts
```diff
@@ -0,0 +1,59 @@
+import { storage } from '#imports'
+import { THEME_MODE_KEY } from './constants/storage-keys'
+
+export type ThemeMode = 'light' | 'dark' | 'system'
+
+let cachedThemeMode: ThemeMode = 'system'
```
mengxi-ream, line 6: shouldn't have theme `cachedThemeMode` and `isInitialized`

because the service worker (background script) will be inactivate after certain time and them these variable will all be unavailable.

You should store all persistent things in storage API, and use Provider to get and set value into storage API.

mengxi-ream (changes requested):
Hi thanks for the PR.

But in the high level, your design is wrong and has the bug.

1. you should not use memory vairable like `cachedTheme` as the reference since they are not available when service worker or content scripts become inactive (sleep)
2. the theme provider has two states `theme`, `themeMode` is a bad design. because you have kind of duplicated state and you should maintain some logic to make sure these two states don't conflict with each other. the good design is only has one state `themeMode` = `light | dark | system` (read from storage API) and another value is derived from this states (e.g. `theme = getResolvedTheme(themeMode)`, rather than a new state. because you just need a derived value from one state, which is the only one source of truth.
3. all content scripts is not affected by your theme setting
CleanShot 2025-11-22 at 21 11 22@2x

I may give you some guidance:

So you should save the themeMode into local storage, read the themeMode from local storage. and in theme provider, you gonna have a `const theme` derived from `themeMode`, rather than a state `[theme, ] = useState(...)`

mengxi-ream:
another hint: see how we design and sync `config` between storage API with react states using `storage-adapter`

https://github.com/mengxi-ream/read-frog/blob/main/src/utils/atoms/config.ts#L8

So your `themeMode` is actually just another `config` that you need sync themeMode between `storage` and React `state`

mengxi-ream:
I think you might be a student, as a student, you already did a good job. But this question does test your understanding of state management and the browser extension service worker sleep mechanism. You can ask me any questions about this PR, or ask the AI to understand these things in deeper

PR #1130, opened 2026-03-16
fix(background): respect site control for blank iframes

YsLtr:
## Summary
- resolve site-control checks for programmatic iframe injection against the nearest real page URL
- prevent temporary about:blank/about:srcdoc documents from bypassing disabled-site rules
- add documentId-based deduplication for iframe injection and cover the URL resolution with a sanitized regression test

## Problem
Programmatic iframe injection used the current frame URL directly. When a site created temporary about:blank or about:srcdoc documents, those URLs did not match the disabled-site patterns, so the extension could still inject selection and host content scripts into frames that belonged to a disabled page.

## Why not skip about:blank/about:srcdoc entirely?
That would be simpler, but it would also drop support for real content that is rendered inside temporary frames. The background iframe injection logic exists specifically to handle dynamic and special iframe cases. Some sites create an about:blank or about:srcdoc frame first and then populate it with the actual page content. Treating every such frame as a non-page would turn this bug into a functional regression.

Instead, this change keeps support for those frames and resolves site control against the nearest real ancestor/top-level page URL.

## Validation
- pnpm exec eslint src/entrypoints/background/iframe-injection.ts src/entrypoints/background/iframe-injection-utils.ts src/entrypoints/background/__tests__/iframe-injection-utils.test.ts
- pnpm exec vitest run src/entrypoints/background/__tests__/iframe-injection-utils.test.ts
- pnpm exec tsc --noEmit
- SKIP_FREE_API=true git push (pre-push hook ran lint, type-check, and tests with the documented free-api skip)

### Before fix
Clip_20260316_213051

There were a large number of requests for selection.css, which severely slowed down the website loading speed.

### After fix
Clip_20260316_213323

The requests for selection.css disappeared, and the website loading speed returned to normal.

---
## Summary by cubic
Fix iframe content script injection to respect disabled-site rules by resolving site control from the nearest real page and applying it to `about:blank`/`about:srcdoc` iframes. Also blocks background injections on disabled sites, deduplicates per-document injections, and adds tests.

- **Bug Fixes**
  - Resolve site-control URL for `about:blank`/`about:srcdoc` by walking ancestor frames; write it to `window` for content scripts; clear the override on navigation, teardown, and invalidation.
  - Guard background injection: target by documentId (fallback to frameId), skip when site is disabled, prevent in-flight duplicates, prune stale frames, and reset per-tab/frame state on navigations and tab removal.
  - Improve logging and failure handling for URL resolution/injection; add tests for URL resolution, background injection behavior, and effective site-control decisions.

<sup>Written for commit 755d47259b051b44269cbed395adc9c695a19f5c. Summary will update on new commits.</sup>


YsLtr:
> ## PR Review: fix(background): respect site control for blank iframes
> ### Overall
> Solid bug fix with clear motivation. The approach of resolving site control against the nearest real ancestor URL is the right call over blanket-skipping `about:blank`/`about:srcdoc`. The extraction of `resolveSiteControlUrl` into a tested utility is clean.
> 
> ### Code Quality
> **iframe-injection-utils.ts** — Clean, focused, testable. The `isSiteControlUrl` type guard and frame-tree walk are straightforward. Good infinite-loop guard with `parentFrameId === currentFrame.frameId`.
> 
> **iframe-injection.ts** — The nested try-catch adds visual complexity but serves distinct purposes (prerequisite resolution vs. script injection), so it's justified. The deduplication logic with `pendingDocumentKeys`/`injectedDocumentKeys` is a nice touch for avoiding redundant injections.
> 
> ### Potential Issues
> 1. **Memory growth in `injectedDocumentKeys`** — Keys are only cleaned on `tabs.onRemoved`. For long-lived tabs with many navigations (SPA with iframes, or sites that continuously create/destroy iframes), this set grows without bound. Consider also cleaning on `browser.webNavigation.onBeforeNavigate` for the top frame (frameId === 0) to clear stale entries when the user navigates within the same tab.
> 2. **`resolveSiteControlUrl` builds a new `Map` on every call** — This is fine for the current call frequency (once per iframe load), but worth noting if this function ever gets called in a hot path. Not a blocker.
> 3. **Missing `documentId` fallback** — When `documentId` is `undefined` (e.g., older Chromium versions), `getDocumentInjectionKey` returns `null` and deduplication is completely skipped. This is acceptable as a graceful degradation, but a brief comment explaining this intentional behavior would help future readers.
> 4. **Race between `pendingDocumentKeys.add` and the async site-control check** — If two `onDOMContentLoaded` events fire for the same document in rapid succession, both could pass the `has()` check before either adds the key. This is unlikely in practice (DOM content loaded fires once per document), but if the listener can fire multiple times for the same `documentId`, the pending guard won't fully deduplicate. Low risk.
> 
> ### Security
> This is a security improvement — it closes a bypass where disabled-site rules were not enforced on `about:blank`/`about:srcdoc` iframes. The fix correctly walks up the frame tree to find the real origin URL.
> 
> One edge case: if _all_ ancestor frames are `about:blank` (no real URL in the chain), `resolveSiteControlUrl` falls back to `frameUrl` (the original `about:blank`). This means `isSiteEnabled("about:blank", config)` gets called, which would return `true` in blacklist mode (since `about:blank` won't match any domain pattern). Whether this is desired depends on the threat model — it's a narrow edge case but worth documenting the decision.
> 
> ### Test Coverage
> The tests for `resolveSiteControlUrl` cover the key scenarios well:
> 
> * Regular URL passthrough
> * Single-level `about:blank` resolution
> * Multi-level walk (`about:srcdoc` → `about:blank` → real URL)
> * Missing frame fallback to top frame
> 
> **Suggestions for additional coverage:**
> 
> * Test where the top frame is also `about:blank` (all-blank chain) to verify the fallback behavior
> * Test with `file://` URLs to ensure they're treated as real site URLs
> * Integration-level test for the deduplication logic (pending/injected sets) would be valuable but is understandably harder to set up
> 
> ### Minor Nits
> * The `for...of` + `startsWith` cleanup pattern in `onRemoved` is O(n) per set. If tab churn is high with many tracked documents, a `Map<tabId, Set<key>>` would be O(1) cleanup. Not a concern at current scale, but something to revisit if the sets grow large.
> 
> ### Verdict
> 🟢 Good taste. Clean extraction, proper fix for a real user-facing bug, and good test coverage on the new utility. The memory growth concern (point 1) is the most actionable item — the rest are minor observations. Nice work.

Applied the stale-key cleanup on top-frame navigation and documented the intentional documentId fallback. I did not change resolveSiteControlUrl because the current call frequency does not justify extra complexity. The reported has()/ add() race is not possible here because that section runs synchronously before the first await.

mengxi-ream:
@codex review this PR

ananaBMaster:
@YsLtr which website you use to test and debug?

YsLtr:
My school's inner course website. Use tronclass as template.

> @YsLtr which website you use to test and debug?

ananaBMaster:
@YsLtr I made some change, could you check if this break anything?

YsLtr:
> @YsLtr I made some change, could you check if this break anything?

I've checked for errors or breaks, there are no more conflict. I've built the code you committed and test on the actual site, no more breaks were found.
Clip_20260319_124554

But Cubic-dev-ai's issue was verified by Codex to exist, and it should be fixed later.

ananaBMaster:
> > @YsLtr I made some change, could you check if this break anything?
> 
> 
> 
> I've checked for errors or breaks, there are no more conflict. I've built the code you committed and test on the actual site, no more breaks were found.
> 
> Clip_20260319_124554
> 
> 
> 
> But Cubic-dev-ai's issue was verified by Codex to exist, and it should be fixed later.

I will fix tomorrow

PR #1242, opened 2026-03-26
fix: storage false value reset and backup delete dialog not showing

kilidoc:
## Type of Changes

- [ ] ✨ New feature (feat)
- [x] 🐛 Bug fix (fix)
- [ ] ✅ Test related (test)
- [ ] 🔄 CI/CD related (ci)
- [ ] 🧠 AI model related (ai)

## Description

- fix(storage): use strict null check (`!== null && !== undefined`) instead of falsy check so boolean `false` is correctly persisted; this caused the analytics toggle to reset to `true` on every settings page reload
- fix(backup): replace `onSelect` with `onClick` on DropdownMenuItem so the AlertDialog is not dismissed before it can open when clicking Delete

### DIFF
--- a/.changeset/late-apples-care.md
+++ b/.changeset/late-apples-care.md
@@ -1,0 +1,5 @@
+---
+"@read-frog/extension": patch
+---
+
+perf(subtitles): decouple AI smart context summary from translation
--- a/.changeset/neat-views-fetch.md
+++ b/.changeset/neat-views-fetch.md
@@ -1,0 +1,5 @@
+---
+"@read-frog/extension": patch
+---
+
+feat(models): update minimax model list and default model
--- a/.changeset/plain-yaks-warn.md
+++ b/.changeset/plain-yaks-warn.md
@@ -1,0 +1,5 @@
+---
+"@read-frog/extension": patch
+---
+
+fix(extension): preserve shared popup close-state behavior in builds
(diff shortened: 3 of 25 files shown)


src/utils/atoms/storage-adapter.ts
```diff
@@ -4,7 +4,7 @@ import { storage } from "#imports"
 export const storageAdapter = {
   async get<T>(key: string, fallback: T, schema: ZodSchema<T>): Promise<T> {
     const value = await storage.getItem<T>(`local:${key}`)
-    if (value) {
+    if (value !== null && value !== undefined) {
       const parsedValue = schema.safeParse(value)
```
ishiko732, line 9: I recommend using `typeof value !== 'undefined'` because it can be optimized by V8.

kilidoc: Thanks for the suggestion! I'm a bit concerned that typeof value !== 'undefined' would let null pass through. Does the storage API guarantee it never returns null? If not, I think we may still need to handle that case.

ishiko732: You need to handle value === '' and typeof value === 'undefined'.

ishiko732: > Thanks for the suggestion! I'm a bit concerned that typeof value !== 'undefined' would let null pass through. Does the storage API guarantee it never returns null? If not, I think we may still need to handle that case.

You need to keep the null check as well. Since this involves checking three cases, you can use:
```ts
![null, '', undefined].some(v => v === value)
```

> Also, null will not pass Zod’s parse validation, so it is safe.

PR #1293, opened 2026-04-08
feat(options): add settings language switcher for options page

alexj11324:
## Type of Changes

- [x] ✨ New feature (feat)
- [x] 🌐 Internationalization (i18n)

## Description

Adds a language switcher dropdown to the options page header, allowing users to override the extension's UI language independently of the browser locale.

**Key changes:**
- New `SettingsLanguageSwitcher` component rendered in the General page header
- New `i18n-override.ts` module that patches `i18n.t()` from `@wxt-dev/i18n` to serve messages from a user-selected locale
- `setupOptionsLocaleOverride()` is called before React rendering in `main.tsx` to ensure all components render with the correct locale
- Supports all 8 locales: en, zh-CN, zh-TW, ja, ko, ru, tr, vi

**Technical approach:**
Instead of monkey-patching `browser.i18n.getMessage` (which can silently fail in Chrome MV3 where native API properties may be non-writable), this patches `i18n.t()` directly — a plain JS object property that is guaranteed to be writable. The locale preference is persisted in `localStorage` and applied on page load via `location.reload()`.

## Related Issue

N/A

## How Has This Been Tested?

- [x] All 964 existing unit tests pass
- [x] TypeScript type-check passes
- [x] ESLint passes

## Checklist

---
## Summary by cubic
Add a language switcher to the Options page header so users can choose the UI language independent of the browser locale. The selected locale loads at startup so the page renders in the correct language.

- **New Features**
  - Added `SettingsLanguageSwitcher` in the General page header; supports en, zh-CN, zh-TW, ja, ko, ru, tr, vi.
  - Patched `i18n.t` from `@wxt-dev/i18n` to serve messages from the selected locale; stored in `localStorage` and applied via `setupOptionsLocaleOverride()` before React renders.
  - Added `headerRight` prop to `PageLayout` to host the switcher in the header.

<sup>Written for commit c996e33cb49bd98359e4405b29f963ce7f0900e6. Summary will update on new commits.</sup>

### DIFF
--- a/src/entrypoints/options/pages/general/index.tsx
+++ b/src/entrypoints/options/pages/general/index.tsx
@@ -1,10 +1,10 @@
 import { i18n } from "#imports"
 import { PageLayout } from "../../components/page-layout"
+import SettingsLanguageSwitcher from "../../components/settings-language-switcher"
 import AppearanceSettings from "./appearance-settings"
 import FeatureProvidersConfig from "./feature-providers-config"
 import LanguageDetectionConfig from "./language-detection-config"
 import SiteControlMode from "./site-control-mode"
-import SettingsLanguageSwitcher from "../../components/settings-language-switcher"
 
 export function GeneralPage() {
   return (


frogGuaGuaGuaGua:
Thanks for the contribution. I'm going to close this PR.

The main reason is architectural rather than just implementation details. This PR achieves runtime language switching by patching WXT/browser i18n behavior inside the options page. We don't want to support selectable UI languages through a partial workaround like this, because it creates a one-off path for the options page, couples us to @wxt-dev/i18n internals, and increases long-term maintenance cost over time.

More importantly, this kind of monkey-patching is likely to become fragile as the codebase evolves. Future upgrades to WXT, @wxt-dev/i18n, or our extension architecture could easily break this behavior in subtle ways, and we don't want to take on that risk for a feature that would need a more complete solution anyway.

If read-frog decides to support user-selectable UI languages in the future, we want to do it with a proper i18n solution and a full refactor across the extension rather than monkey-patching the current i18n layer.

Thanks again for exploring the problem space.

PR #1315, opened 2026-04-12
fix(selection): close language selector dropdown after selection

pzzyf:
Fix target language selector dropdown not closing after selecting a language 

---
## Summary by cubic
Fixes the target language selector in the selection toolbar so the dropdown closes after choosing a language. Improves UX by dismissing the popover on selection and ignoring no-op choices.

- **Bug Fixes**
  - Control the `Combobox` open state with local state and `onOpenChange` from `@base-ui/react`.
  - Close the dropdown in `onValueChange` and update `targetCode`.
  - Also close on null, `auto`, or same language selections to avoid no-op updates.

<sup>Written for commit 2cd37481c95b7dde290d1152f6fa03963e992f30. Summary will update on new commits.</sup>

### DIFF
--- a/src/entrypoints/selection.content/selection-toolbar/translate-button/target-language-selector.tsx
+++ b/src/entrypoints/selection.content/selection-toolbar/translate-button/target-language-selector.tsx
@@ -1,8 +1,9 @@
+import type { LanguageItem } from "@/components/language-combobox-options"
 import { i18n } from "#imports"
 import { Combobox as ComboboxPrimitive } from "@base-ui/react"
 import { IconChevronDown } from "@tabler/icons-react"
 import { useAtom } from "jotai"
-import { useMemo } from "react"
+import { useCallback, useMemo, useState } from "react"
 import { filterLanguage, getTargetLanguageItems } from "@/components/language-combobox-options"
 import { Button } from "@/components/ui/base-ui/button"
 import {
@@ -25,17 +26,23 @@ export function TargetLanguageSelector() {
     [language.targetCode, languageItems],
   )
   const title = currentItem?.label ?? i18n.t("side.targetLang")
+  const [open, setOpen] = useState(false)
+  const onValueChange = useCallback((item: LanguageItem | null) => {
+    if (!item || item.value === "auto" || item.value === language.targetCode) {
+      setOpen(false)
+      return
+    }
+
+    void setLanguage({ targetCode: item.value })
+    setOpen(false)
+  }, [language.targetCode, setLanguage])
 
   return (
     <Combobox
+      open={open}
+      onOpenChange={setOpen}
       value={currentItem}
-      onValueChange={(item) => {
-        if (!item || item.value === "auto" || item.value === language.targetCode) {
-          return
-        }
-
-        void setLanguage({ targetCode: item.value })
-      }}
+      onValueChange={onValueChange}
       items={languageItems}
       filter={filterLanguage}
       autoHighlight


mengxi-ream:
I believe this has been addressed in 1.32.1

pzzyf:
> I believe this has been addressed in 1.32.1

The upstream main branch (latest) still has no open state control in                 
  target-language-selector.tsx. The onValueChange callback only calls setLanguage without
   closing the dropdown. This issue is not fixed in 1.32.2. You can verify by checking   
  the file at src/entrypoints/selection.content/selection-toolbar/translate-button/target
  -language-selector.tsx on the main branch. Or is it designed to stay open              
  intentionally?

mengxi-ream:
We don't have version 1.32.2. where did you get that?

 and I can't reproduce this in 1.32.1


https://github.com/user-attachments/assets/31effd02-2766-4efe-b019-fd896415a9b8

mengxi-ream:
@frogGuaGuaGuaGua also verify this Please

mengxi-ream:
> > I believe this has been addressed in 1.32.1
> 
> 
> 
> The upstream main branch (latest) still has no open state control in                 
> 
>   target-language-selector.tsx. The onValueChange callback only calls setLanguage without
> 
>    closing the dropdown. This issue is not fixed in 1.32.2. You can verify by checking   
> 
>   the file at src/entrypoints/selection.content/selection-toolbar/translate-button/target
> 
>   -language-selector.tsx on the main branch. Or is it designed to stay open              
> 
>   intentionally? 

There is a concept called uncontrolled state, please check base ui doc

mengxi-ream:
@frogGuaGuaGuaGua explain what is uncontrolled states in this context

frogGuaGuaGuaGua:
I checked current main plus the Base UI docs.

In this component, the selected value is already controlled by app state (`value={currentItem}` + `onValueChange={...}`), but the popup open state is not controlled because we do not pass `open` / `onOpenChange`.

That is what "uncontrolled" means in this context: the Combobox popup manages its own open/close state internally.

Base UI docs explicitly describe this behavior:
- Combobox docs: `defaultOpen` is for the uncontrolled popup state, and `open` is only needed when you want a controlled popup.
- Customization docs: components are uncontrolled by default and manage their own state internally.

So the absence of local `open` state in `target-language-selector.tsx` is not evidence by itself that the dropdown should stay open. In the default/uncontrolled mode, Base UI is expected to close it on selection by itself.

So I would interpret this PR as changing the popup from uncontrolled open state to controlled open state and then forcing `setOpen(false)` after selection, rather than fixing a missing open-state control bug in the abstract.

If there is a real repro on current main where it stays open after selection, that repro would be the important evidence. But from the code alone, Mengxi is right that "no explicit open state" != "bug", because Base UI already owns that state in uncontrolled mode.

Docs referenced:
https://base-ui.com/react/components/combobox
https://base-ui.com/react/handbook/customization

pzzyf:
> We don't have version 1.32.2. where did you get that?
> 
> and I can't reproduce this in 1.32.1
> 
>  CleanShot.2026-04-12.at.01.30.18.mp4


https://github.com/user-attachments/assets/fa40235d-2b62-44b3-b6df-a0134b40bd71

mengxi-ream:
> > We don't have version 1.32.2. where did you get that?
> 
> > 
> 
> > and I can't reproduce this in 1.32.1
> 
> > 
> 
> >  CleanShot.2026-04-12.at.01.30.18.mp4
> 
> 
> 
> 
> 
> https://github.com/user-attachments/assets/fa40235d-2b62-44b3-b6df-a0134b40bd71
> 
> 
> 
> 

This version is 1.31.2 which is previous version of 1.32.0

PR #1401, opened 2026-04-27
feat: add app language setting

byte92:
## Type of Changes

- [ ] ✨ New feature (feat)
- [ ] 🐛 Bug fix (fix)
- [ ] ✅ Test related (test)
- [ ] 🔄 CI/CD related (ci)
- [x] 🌐 Internationalization (i18n)
- [ ] 🧠 AI model related (ai)

## Description

Add an app language setting under General settings so users can choose the Read Frog UI language explicitly instead of only relying on the browser locale.

This PR adds:

- A new App Language selector in General settings
- Locale preference storage with a "Follow browser language" default
- Runtime i18n message override for options, popup, translation hub, side content, selection toolbar, host toast, and subtitles UI
- Localized labels for all existing supported app locales

## Related Issue

N/A

## How Has This Been Tested?

- [ ] Added unit tests

Tested with:

```bash
pnpm lint
SKIP_FREE_API=true pnpm exec vitest run src/utils/constants/__tests__/model.test.ts src/utils/config/__tests__/init.test.ts
WXT_SKIP_ENV_VALIDATION=true pnpm build
```

Result: all commands above passed.

Note: `pnpm type-check` currently fails because `src/utils/host/translate/webpage-context.ts` cannot resolve `defuddle/full`, which appears unrelated to this change.

## Screenshots

Not included.

## Checklist

- [ ] I have updated the documentation accordingly if necessary

## Additional Information

The default behavior remains unchanged: app language follows the browser language unless the user selects a specific locale.


mengxi-ream:
Thanks for the contribution and for taking the time to add an explicit app language setting.

I am going to close this PR for now because the current implementation works by monkey-patching the runtime i18n layer instead of using a supported browser extension i18n API. Browser extension i18n does not provide a stable runtime API for switching the UI locale, so this patch has to override WXT i18n.t, load _locales/<locale>/messages.json manually, and reimplement parts of message lookup, substitutions, plural handling, and fallback behavior.

That creates a fairly high maintenance risk: if WXT changes its i18n runtime API, generated message format, argument semantics, fallback behavior, or if browser extension resource access rules differ across Chrome/Firefox/content-script contexts, this can silently break or produce mixed-language UI. It also does not cover every i18n surface equally, such as browser-managed extension strings and background/context-menu text.

The feature request itself is valid, but we should revisit it with a more explicit and maintainable app-localization layer rather than patching the existing i18n runtime globally. Thanks again for the work here.

Issue #1403, opened 2026-04-27
[FEATURE] show translation status on popup icon


mengxi-ream:
### Before you begin

- [x] I have searched existing issues to avoid duplicates
- [ ] I am willing to contribute a PR for this feature

### Is your feature request related to a problem? Please describe.

show translation status on popup icon

### Describe the solution you'd like

show translation status on popup icon

### Which area(s) would this feature affect? (Select all that apply)

Browser Extension

TrivCodez:
Hey, I'd like to work on this feature! It seems straightforward.

PR #1404, opened 2026-04-27
feat: show translation status on popup icon

TrivCodez:
## What
Adds a visual translation status indicator to the popup header showing whether the current page is translated.

## Why
Addresses issue #1403 - users want to quickly see translation status without having to check the page manually.

## Changes
- Added `TranslationStatusIndicator` component that reads from existing `isPageTranslatedAtom`
- Integrated indicator into popup header next to other action buttons
- Shows green dot + "Translated" text when page is translated, gray dot + "Not Translated" otherwise
- Uses existing i18n system for localization

## Testing
- [x] Component renders correctly in popup
- [x] Status updates when page translation state changes
- [x] Uses theme-aware styling (works in both light/dark modes)

## Notes
- This is a UI-only change that leverages existing state management
- No breaking changes to existing functionality
- Ready for review


mengxi-ream (changes requested):
CleanShot 2026-04-27 at 10 04 25@2x

we mean status on this icon

PR #1529, opened 2026-05-16
docs: clarify LM Studio local provider setup

frogGuaGuaGuaGua:
## Summary
- Add LM Studio guidance to the Ollama provider docs in both Chinese and English
- Clarify that LM Studio should use Custom Provider / OpenAI-compatible with `http://localhost:1234/v1`
- Document API key behavior and model identifier expectations so users do not configure LM Studio as Ollama

## Notes
- The current `main` branch does not contain the website docs source. This PR targets `production`, where `apps/website/content/docs/providers/ollama*.mdx` currently lives.

## Test Plan
- [x] `pnpm install --frozen-lockfile`
- [x] `git diff --check`
- [x] `pnpm --filter @read-frog/website build`
- [x] pre-push hook: `syncpack:lint`, `type-check`, `build`, `test`

Refs #1522


frogGuaGuaGuaGua:
CI note: the failing **Score contributor trust** check fails before it reads this PR's docs changes. The workflow is `pull_request_target` from the default branch and checks out `github.event.pull_request.base.sha`; because this PR targets `production`, that checkout is the current `production` tree, which does not contain `.github/scripts/contributor-trust/run.js`.

The docs change itself was validated locally and by the pre-push hook:

- `pnpm --filter @read-frog/website build`
- pre-push `syncpack:lint`
- pre-push `type-check`
- pre-push `build`
- pre-push `test`

I targeted `production` because the current `main` branch does not contain the website docs source for this page. If there is a newer canonical docs branch/source path, this PR can be retargeted or the same MDX change can be moved there.

PR #1579, opened 2026-05-23
fix(translation): force block X/Twitter tweet text via DOM rules

irandoku:
## Problem

X/Twitter tweet text is rendered under `[data-testid="tweetText"]` with nested inline elements. In bilingual translation mode, the translation and original text appear merged without clear paragraph separation.

## Root Cause

`unwrapDeepestOnlyHTMLChild()` (called inside `translateNodesBilingualMode`) unwraps `targetNode` from the `tweetText` container to its inner `<span>` when the container has exactly one HTML child and no text children.

After unwrapping, `isCustomForceBlockTranslation(targetNode)` checks the inner `<span>` against the selector — which does not match `[data-testid="tweetText"]` — so the force-block flag is never set, and `addInlineTranslation` is called instead of `addBlockTranslation`.

## Why `[data-testid="tweetText"] *` broke layout

The `*` descendant selector matched every descendant inside `tweetText`, causing each inline child to be force-blocked individually. This turned every word/phrase into a separate block with its own `<br>`, breaking tweet layout.

## Fix: `:only-child` selector

```typescript
"x.com": [
  "[data-testid=\"tweetText\"]",
  "[data-testid=\"tweetText\"] > span:only-child",
],
"twitter.com": [
  "[data-testid=\"tweetText\"]",
  "[data-testid=\"tweetText\"] > span:only-child",
],
```

- `[data-testid="tweetText"]` — matches the container itself (handles the case where `targetNode` stays as the container)
- `[data-testid="tweetText"] > span:only-child` — matches the inner `<span>` only when it is the sole child of the container (handles the `unwrapDeepestOnlyHTMLChild` case)

The `:only-child` pseudo-class ensures this selector only activates when `unwrapDeepestOnlyHTMLChild` would unwrap to the inner span. For tweets with multiple children (e.g. with images, emojis, or multiple text segments), `:only-child` does not match, and the container is translated as a whole — avoiding the layout breakage from `*`.

## Changed files

- `src/utils/constants/dom-rules.ts` (+8 lines) — add X/Twitter entries to `CUSTOM_FORCE_BLOCK_TRANSLATION_SELECTOR_MAP`

## Verification

- [x] Pre-push checks: lint, type-check, test suite all pass
- [x] Local Chrome MV3 build installed and tested on X/Twitter
- [x] Random blind test on multiple X/Twitter posts — original and translated text correctly separated with `<br>`
- [x] Verified with browser DevTools: translation wrapper uses `read-frog-translated-block-content` (not `inline-content`)
- [x] Only changes DOM rules file — no changes to walker, filter, or translation insertion behavior

### DIFF
--- a/src/utils/constants/dom-rules.ts
+++ b/src/utils/constants/dom-rules.ts
@@ -168,4 +168,12 @@ export const CUSTOM_FORCE_BLOCK_TRANSLATION_SELECTOR_MAP: Record<string, string[
   "www.youtube.com": [
     "yt-attributed-string > span",
   ],
+  "x.com": [
+    "[data-testid=\"tweetText\"]",
+    "[data-testid=\"tweetText\"] > span:only-child",
+  ],
+  "twitter.com": [
+    "[data-testid=\"tweetText\"]",
+    "[data-testid=\"tweetText\"] > span:only-child",
+  ],
 }


irandoku:
Screenshot of the issue on an @elonmusk tweet — English original and Chinese translation merged without paragraph separation:

irandoku:
Capture_2026-05-24_03-22-53_AM

Screenshot of the issue: English original and Chinese translation merged without paragraph separation on X/Twitter. See PR description for full analysis.

mengxi-ream:
The only fix way to accept is to change the dom rule file. Don't change the walking algorithm which break other sites style

irandoku:
You were right. I reworked this PR so it only changes the DOM rule file behavior surface.

The nested tweet text case can be handled by expanding the X/Twitter selector to include descendants:

- `[data-testid="tweetText"]`
- `[data-testid="tweetText"] *`

This avoids changing `isCustomForceBlockTranslation`, the walker, or translation insertion behavior. I also verified it with a local Chrome MV3 build on X/Twitter; bilingual translation now separates the original tweet text and translated text correctly.

mengxi-ream (changes requested):
CleanShot 2026-05-25 at 23 12 09@2x
you rule breaks the style

irandoku:
Updated this PR with a new approach that fixes the issue without breaking tweet layout.

**Key finding:** The root cause is `unwrapDeepestOnlyHTMLChild()` — it unwraps `targetNode` from the `tweetText` container to its inner `<span>` (when the container has exactly one HTML child). After unwrapping, `[data-testid="tweetText"]` no longer matches the targetNode, so force-block is never triggered.

**New selector:** Instead of `*` (which matched all descendants and broke layout), I use `[data-testid="tweetText"] > span:only-child` — this only matches the inner span when it is the sole child, which is exactly the case where `unwrapDeepestOnlyHTMLChild` unwraps to it. For tweets with multiple children (images, emojis, etc.), `:only-child` does not match, so there is no layout breakage.

**Changed files:** Only `src/utils/constants/dom-rules.ts` (+8 lines). No changes to walker, filter, or translation insertion.

Tested with random blind tests on multiple X/Twitter posts — all working correctly.

mengxi-ream:
Because this PR introduced some new bugs, we don't fully trust it and have decided not to merge it for now.

Instead, we will launch a comprehensive rule system by the end of this year, with a particular focus on improving the translation experience for Twitter.

Thank you for the efforts!

irandoku:
Thanks for the clear explanation @mengxi-ream. Totally understand — quality and stability come first, and a comprehensive rule system for Twitter translation sounds like the right long-term direction. Happy to help test or contribute when that work kicks off. Thanks for the review!

PR #1666, opened 2026-06-10
feat: add API key rotation fallback

Opeple0627:
## Summary
- allow comma-separated API keys for API providers
- add sequential/random key rotation and configurable failed-key cooldown
- immediately fall back to another key for API/LLM translation, DeepL/DeepLX, generateText, summaries, segmentation, and pre-output stream failures

## Verification
- pnpm exec tsc --noEmit
- pnpm exec eslint on affected files
- pnpm exec vitest run src/entrypoints/background/__tests__/background-stream.test.ts src/entrypoints/background/__tests__/llm-generate-text.test.ts src/entrypoints/background/__tests__/translation-queues.test.ts src/utils/providers/__tests__/api-key-rotation.test.ts src/utils/providers/__tests__/model.test.ts src/utils/host/translate/api/__tests__/ai.test.ts src/utils/host/translate/api/__tests__/deepl.test.ts
- pnpm exec vitest run src/utils/host/translate/ui/__tests__/spinner.test.ts

Note: full `pnpm exec vitest run` still has the existing spinner test timing/mock-isolation failure when run with the full suite, while the spinner test passes standalone.

### DIFF
--- a/.changeset/api-key-rotation.md
+++ b/.changeset/api-key-rotation.md
@@ -1,0 +1,5 @@
+---
+"@read-frog/extension": patch
+---
+
+Add comma-separated API key rotation with sequential or random selection, cooldown for failed keys, and immediate fallback to another configured key.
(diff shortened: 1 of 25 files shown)


mengxi-ream:
The rotate function should be implemented by users setting up their own proxy layer. We will not maintain such extensive proxy code within the software unless the number of permanent developers increases in the future.

PR #1749, opened 2026-06-27
feat(subtitles): add official bilingual streaming captions

semantic-craft:
## Summary

This PR adds a streaming bilingual subtitle path for Netflix-style players. When both official source and target subtitle tracks are available, Read Frog now renders the official original subtitle together with the official localized subtitle instead of always falling back to machine translation.

Concretely, this adds support for:

- discovering official subtitle tracks from streaming manifests and late JSON/XHR/fetch responses
- pairing official English captions with official target-language subtitles, including Chinese Traditional on Netflix
- aligning English cues to the target-language cue timing with maximum-overlap matching, so cue splits do not truncate or duplicate lines
- preserving the user's selected native Netflix subtitle track instead of forcing English when a usable target subtitle is already enabled
- falling back to native/live subtitle capture when official tracks are unavailable
- parsing Netflix TTML tick-based cue timing

## Product advantage

The main product difference is that this is not just another machine-translated subtitle overlay.

A well-known extension such as Immersive Translate usually solves video bilingual subtitles by taking the selected English subtitle and generating a machine translation. That is useful when no official translation exists, but it cannot provide a perfect official-to-official comparison: the localized line is generated by the translation provider, not taken from the streaming platform's official subtitle asset.

Read Frog can now do better on Netflix when both tracks are present. It keeps the official Chinese Traditional subtitle selected in Netflix, captures the official English caption track separately, and renders them together as a matched bilingual pair. In other words, the viewer gets official English above official Chinese, rather than English plus a machine-generated Chinese translation. This is a stronger experience for language learning because users can compare the original wording with the platform-approved localization.

## Why this was hard

Official subtitle tracks are not segmented the same way across languages. Netflix may split one English sentence across multiple cues while the Chinese subtitle uses a single cue, or the reverse. A simple timestamp overlap would either truncate the English line, duplicate translations, or attach a source cue to the wrong localized line.

The implementation addresses this by using the target-language subtitle timing as the display baseline and assigning each source cue to the target cue with the largest time overlap. That lets Read Frog display complete official English lines above the corresponding official Chinese subtitle, even when the underlying tracks use different cue boundaries.

## Implementation notes

- The page-world interceptor now scans nested manifest-like JSON for timed text downloadables and replays cached tracks when the subtitles runtime starts later.
- Official track selection filters out forced/image/watch-route noise and prefers caption-like English tracks over original-audio metadata tracks.
- Official bilingual subtitles are pre-segmented and bypass machine translation when a valid official source/target pair is available.
- Native subtitle fallback timing now updates live cue end times so fallback subtitles do not linger after the spoken line has ended.

## Validation

- `pnpm type-check`
- `pnpm exec eslint`
- `SKIP_FREE_API=true pnpm vitest run src/utils/subtitles/__tests__/streaming-fetcher.test.ts src/entrypoints/interceptor.content/__tests__/streaming-subtitles-interceptor.test.ts src/entrypoints/subtitles.content/__tests__/universal-adapter.test.ts src/entrypoints/subtitles.content/__tests__/subtitles-scheduler.test.ts`
- `SKIP_FREE_API=true pnpm test` — 166 files passed, 1417 tests passed
- `WXT_SKIP_ENV_VALIDATION=true pnpm exec wxt build`
- Local Chrome extension validation on Netflix with official Chinese Traditional subtitles selected; confirmed official English captions render above the official Chinese subtitle line without truncating the English cue.


semantic-craft:
Withdrawing this official-track approach in favor of #1758.

This PR's goal was official Chinese subtitle + official English subtitle comparison. I still want to support that later, but the full official-track alignment path adds enough parsing, matching, and provider-specific code that compressing it under the current small-PR / sub-1000-line review target would make the implementation incomplete.

The replacement PR first lands the smaller, repeatedly locally tested streaming-caption path: capture the player-rendered English cue and translate it through Read Frog's existing pipeline, similar in user-facing behavior to Immersive Translate. That generic adapter shape has been tested locally on Netflix and HBO Max; #1758 only enables Netflix so the diff stays small and reviewable.

PR #1757, opened 2026-06-28
feat(subtitles): add streaming subtitle adapter framework

semantic-craft:
## Summary

This supersedes #1751 with the systematic adapter shape requested there: streaming subtitle support now goes through one shared pipeline plus a site adapter/registry entry, with Netflix as the first adapter.

Netflix has two paths:
- when Netflix exposes subtitle download tracks in page JSON, Read Frog parses and aligns those official tracks;
- on the current cadmium playback path, where subtitle assets are not directly replayable from the content script, Read Frog falls back to a MAIN-world player/DOM live capture bridge and then uses the existing translation pipeline.

The MAIN-world bridge matters because Netflix's player API is not visible from the extension isolated world. The content script now coordinates through `postMessage`; the interceptor does the player track switch and live cue capture where `window.netflix` actually exists.

## Size / Review

Final diff: 991 insertions / 27 deletions, kept under the 1000-line threshold. HBO stays out of this PR.

## Real Browser Evidence

Verified in the user's Chrome profile on `https://www.netflix.com/watch/81613638` after reloading the unpacked extension:
- Read Frog switch stays ON;
- Netflix player track changes from `zh-Hant` to English CC (`bcp47: en`, `rawTrackType: CLOSEDCAPTIONS`);
- Netflix native timed-text DOM is hidden by Read Frog but continues updating with English cue text;
- Read Frog shadow DOM receives source + translated cue text, e.g. `Probably around the time Emily and I set our boat out on the lake.` plus the Chinese translation.

## Tests

- `pnpm type-check`
- `pnpm exec eslint ...` on touched subtitle/interceptor files
- `WXT_SKIP_ENV_VALIDATION=true pnpm build`
- `SKIP_FREE_API=true pnpm exec vitest run` targeted subtitle tests: 5 files / 85 tests
- pre-push full extension checks: lint, type-check, vitest 163 files / 1400 tests

Closes #1751.


semantic-craft:
Thanks @codex — both points addressed in `6609c86d`:

1. **Netflix watch navigation** (`index.tsx`): removed the ~20s timeout that detached the URL-change listener. The content script now keeps listening for the SPA navigation to `/watch/…` for as long as the page is open, so users who browse a while before picking a title still get the subtitle UI without a reload. (For Netflix, readiness only changes via navigation, so a bounded poll was unnecessary — the listener alone is correct and simpler.)

2. **getVideoId on browse paths** (`netflix/config.ts`): `getVideoId` now returns the `/watch/<id>` id and `null` on browse/details pages. Browse navigations no longer trigger a reset that nulls the session id, so `videoIdChanged` correctly fires when the user opens the next title.

Issue #1369, opened 2026-04-23
[BUG] Options button not working


yaki6:
### Before you report

- [x] I have searched existing issues to avoid duplicates
- [ ] I have tried with the latest available version (or can reproduce on latest)

### Current vs. Expected behavior

Click Options or the gear should leave to the settings page, it's not working

<img width="303" height="245" alt="Image" src="https://github.com/user-attachments/assets/ccecfb4f-3a97-450b-ab22-3def54f94bd4" />

### To Reproduce

1. Click on the extention 
2. Click on gear
3. It's not opening up the setting page

### Which area(s) are affected? (Select all that apply)

Browser Extension

### Provide environment information

```bash
Chrome store 1.32.2
Browser- Arc Version 1.143.2 (79250)
MacOs Tahoe 26.1
```

mengxi-ream:
fixed in https://github.com/mengxi-ream/read-frog/pull/1373

mengxi-ream:
The reason is that Arc doesn't support `browser.runtime.openOptionsPage` api. The fix will release in next version 1.32.3

PR #1777, opened 2026-07-05
fix(extension): open settings via runtime options page

### LINKED ISSUE #1763: [BUG] steam 陪读蛙无法进入 setting 界面
### Before you report

在steam装陪读蛙，无法点击齿轮进入设置界面

### To Reproduce

如上

Browser Extension

_No response_

```bash
1
```

### Additional context

_No response_

lovewave02:
## Summary
- use `browser.runtime.openOptionsPage()` for the default settings entrypoint
- keep routed options-page links on the existing tab/hash fallback path
- add regression coverage for the runtime opener and fallback path

Closes #1763

## Tests
- `SKIP_FREE_API=true pnpm test src/utils/__tests__/navigation.test.ts`
- `pnpm exec eslint src/utils/navigation.ts src/utils/__tests__/navigation.test.ts`
- `SKIP_FREE_API=true pnpm type-check`
- pre-push hook: `nx run @read-frog/extension:lint`, `type-check`, `test` (164 files / 1411 tests passed)


mengxi-ream:
@taiiiyang could you help to review this, also on Windows? maybe steam on MacOS and Windows are slightly different on API.

mengxi-ream:
@taiiiyang Could you please review the latest revision and, if possible, help verify it on Steam?

The implementation now intentionally prefers `browser.tabs.create()` and only falls back to `browser.runtime.openOptionsPage()` when tab creation rejects. We need to keep `tabs.create()` as the primary path because Arc previously could not open the settings page through `runtime.openOptionsPage()` (#1369); that regression was fixed by opening the extension options URL directly in #1373.

I cannot validate the Steam behavior locally on macOS because the Steam client does not appear to provide a way to install or load the extension. The key thing to verify on Steam is whether `browser.tabs.create()` rejects and the runtime fallback successfully opens settings. If `tabs.create()` resolves without showing a page, a catch-based fallback will not be triggered.

taiiiyang:
œœ

> @taiiiyang Could you please review the latest revision and, if possible, help verify it on Steam?
> 
> The implementation now intentionally prefers `browser.tabs.create()` and only falls back to `browser.runtime.openOptionsPage()` when tab creation rejects. We need to keep `tabs.create()` as the primary path because Arc previously could not open the settings page through `runtime.openOptionsPage()` (#1369); that regression was fixed by opening the extension options URL directly in #1373.
> 
> I cannot validate the Steam behavior locally on macOS because the Steam client does not appear to provide a way to install or load the extension. The key thing to verify on Steam is whether `browser.tabs.create()` rejects and the runtime fallback successfully opens settings. If `tabs.create()` resolves without showing a page, a catch-based fallback will not be triggered.

Verified on Steam for both Mac and Windows. 

`browser.runtime.openOptionsPage() ` can properly open the options page.

taiiiyang (approved):

PR #1909, opened 2026-07-19
feat(translate): add vocabulary mode with inline word glosses

s546126:
## Summary

Adds a third translation mode, **vocabulary**, that keeps the original text and annotates words beyond the user's familiar word-frequency rank with an inline parenthesized gloss, e.g. `discover(发现)`.

- Rare-word detection is fully local: a frequency word list from [`most-common-words-by-language`](https://www.npmjs.com/package/most-common-words-by-language) (MIT, Google Web Trillion Word Corpus ranking) plus a user-maintained known-word list. No LLM call is made to decide *which* words are rare — the configured provider only generates glosses for the words already detected, batched and cached in IndexedDB (`wordGlossCache`).
- Inflected forms of common words (e.g. "walked", "studying") are resolved against their base form via a small rule-based candidate generator before falling back to the LLM, so they aren't misjudged as rare just because the inflected form itself ranks lower in the frequency list than its lemma.
- Annotations are written through the existing `translationOnly` in-place text-swap machinery (`planInPlaceTextSwap` / `applyInPlaceTextSwap` / the anchor-state registry), so restore-on-toggle, cleanup, and mid-flight staleness guards behave identically to the other modes — no new DOM lifecycle code was needed for that part.
- Right-click an annotated word → "Mark as known word" (menu item only shown in vocabulary mode) to exclude it from future annotation; the current tab re-walks immediately so the annotation disappears without a manual translate toggle.
- `familiarWordRank` is configurable in Options → Translation (default 5000).

## Why not extend `translationOnly`'s HTML-alignment path instead

`translation-text-swap.ts`'s alignment logic is built for a provider returning a full translated HTML tree that gets paired back against the source DOM. Vocabulary mode's "translation" is assembled locally (original text + inserted glosses) rather than returned by a provider, so it skips that alignment entirely and constructs the swap pairs directly — which also means alignment can't fail here (no fallback-to-detached-wrapper path is reachable).

## Follow-up

An inline "remove gloss" button in the selection toolbar (in addition to the right-click menu entry) is coming in a follow-up PR stacked on this one, to keep this PR's diff focused and under the repo's automated contributor-trust line-count threshold.

## Testing

- `pnpm type-check` — clean
- Focused test run over all changed areas (vocabulary logic, context-menu, migration, translation-modes) — 221/221 passed
- End-to-end verified with a Puppeteer harness driving the built extension in real Chrome against a mock OpenAI-compatible endpoint: annotation, request batching (2 LLM calls for a 3-paragraph fixture, not one per word), in-place swap (anchors present, no wrapper leaks), and full restore-on-toggle (byte-identical to source) all behaved correctly.

## Changed/added areas

- `src/types/config/translate.ts` — `TRANSLATION_MODES` + `vocabularyConfigSchema`
- `src/utils/host/translate/core/translation-modes.ts` — `translateNodesVocabularyMode`, dispatch branch
- `src/entrypoints/background/vocabulary.ts` — gloss batching/caching, known-words message handlers
- `src/entrypoints/background/context-menu.ts` — "mark as known word" menu item
- `src/utils/vocabulary/` — rare-word detection, lemma candidates; frequency list sourced from `most-common-words-by-language` (no data vendored into this repo)
- `src/utils/db/dexie/` — `knownWords` and `wordGlossCache` tables (schema v5)
- `src/utils/config/migration-scripts/v086-to-v087.ts`
- Options UI (`vocabulary-config.tsx`), popup mode selector (now 3-state), i18n (en/zh-CN)

A changeset is included (`.changeset/vocabulary-gloss-mode.md`).


mengxi-ream:
Hi, we really want this feature, but this feature is too complicated to be done just in one PR.
We will implement this shortly in the next few months. Thank you for your PR, I will close this first.

But if you can share us a demo that inspire us how to design this feature in the future. Very welcome to attach that demo here, and we will save it into our internal document for reference

s546126:
Hi @mengxi-ream, thanks for the kind reply — totally understand the concern about landing this in one PR.

A fun detail: the screenshot below is your comment above, rendered by my working build — `complicated(复杂的)`, `shortly(不久)` and `inspire(激发)` were all annotated live by the vocabulary mode. The logic is already running end-to-end, so I'm sharing it as design input.

**Where the idea comes from:** [Relingo](https://relingo.net) does this well, but it's closed-source and can't point at custom OpenAI/Anthropic-compatible endpoints, which is exactly what Read Frog is great at.

**Design notes from my implementation, for your internal doc:**

1. **Rare-word detection is fully local and deterministic.** A frequency-ranked word list plus a user threshold decides which words are rare. The LLM never makes that call — it only glosses the detected words, batched per paragraph and cached in IndexedDB. This keeps token cost low and results stable.
2. **User-maintained known-word blacklist.** Right-click an annotated word or use a selection-toolbar button → mark as known → the page re-annotates immediately. A whitelist of always-gloss words would be the natural complement for exam prep vocab.
3. **Preset levels instead of a raw number.** Ship a few built-in tiers such as top-3k / top-5k / top-8k frequency or CET-4 / CET-6 / CEFR-style lists, so users pick a level and keep the threshold slider for fine-tuning.
4. **Reuse the existing `translationOnly` in-place swap machinery.** The annotated text is just original text plus inserted glosses, assembled locally — so restore-on-toggle, SPA navigation and cleanup all come for free. No new DOM lifecycle code needed.
5. **A small rule-based lemmatizer** maps inflections like `walked` and `studying` back to their base forms before the rarity judgement, otherwise common words get misjudged as rare.

My working branch stays available at [`s546126/read-frog:feat/vocabulary-gloss-mode`](https://github.com/s546126/read-frog/tree/feat/vocabulary-gloss-mode) — feel free to lift anything from it, and I'm glad to help split it into smaller PRs when you start designing this.
Screenshot 2026-07-20 at 9 08 21 AM
Screenshot 2026-07-20 at 9 10 17 AM

PR #1937, opened 2026-07-22
fix(providers): load Gemini models dynamically to keep the list current

### LINKED ISSUE #1938: [BUG] Latest Gemini models unusable: hardcoded model list is outdated and requests fail with "Request contains an invalid argument"

**Current:**

1. The Gemini model dropdown is a hardcoded list, so newly released models (e.g. `gemini-3.6-flash`) never show up.
2. Entering them as a custom model doesn't help either — the request fails with **"Request contains an invalid argument"**.
3. Even some models that *are* in the list fail the same way (e.g. `gemini-3.5-flash-lite`), because the extension hardcodes per-version `thinkingConfig` / sampling parameters that newer Gemini models reject.

**Expected:** newly released Gemini models are selectable and translation requests succeed.

Test connection fails with "Request contains an invalid argument"

### To Reproduce

1. Options → API Providers → Gemini, enter a valid API key
2. Check "Enter the name of the custom model" and type `gemini-3.6-flash`
3. Click Test Connection → ❌ Failed, toast shows "Request contains an invalid argument"

### Which area(s) are affected?

Browser Extension

### Extension version

1.42.2

### Environment

```bash
- OS: Linux
- Browser: Chrome (latest)
- Extension: 1.42.2
```

### Additional context

Root cause: every new Gemini generation requires a manual update of the hardcoded model list and its per-version thinking/sampling overrides, so the extension breaks each time Google ships new models.

I've opened #1937, which fixes this by fetching the model list dynamically from Google's `/models` API and dropping the hardcoded overrides, so future model releases work without code changes.

xuing:
## Type of Changes

- [x] 🐛 Bug fix (fix)

### Problems

1. **The hardcoded Gemini model list was outdated.** Every time Google ships a new model, the extension needs a code change and a release before users can select it.
2. **The newest models were unusable** — e.g. `gemini-3.6-flash` was not in the list at all.
3. **`gemini-3.5-flash-lite` failed at request time**, because the extension hardcoded per-version `thinkingConfig` provider options (`thinkingBudget: 0` / `thinkingLevel: "minimal"`) and sampling parameters that newer Gemini models reject.

### Fix — stop hardcoding, ask the API instead

Rather than patching the static list again (which would break again with the next model generation), this PR makes the Google provider self-updating:

- **Dynamic model discovery**: the model selector now fetches the available models from Google's official [`GET /models`](https://ai.google.dev/api/models#method:-models.list) endpoint (paginated, filtered to `gemini-*` models that support `generateContent`), so newly released models show up automatically. Results are cached for 1 hour via TanStack Query. The API key is debounced and only a truncated SHA-256 fingerprint of it is used in the query cache key, so the key itself never enters the cache.
- **Offline fallback**: the static list is reduced to the three versionless aliases Google keeps up to date themselves (`gemini-flash-lite-latest`, `gemini-flash-latest`, `gemini-pro-latest`), used when the dynamic fetch is unavailable. The default model is now `gemini-flash-lite-latest`.
- **No more per-version thinking hacks**: the hardcoded Gemini `thinkingConfig` patterns are removed and the Google default reasoning is now `provider-default`, letting the Gemini API pick the correct thinking behavior for each model instead of us guessing per version.
- **Strip unsupported sampling params**: Google requests go through a `wrapLanguageModel` middleware that drops `temperature` / `topP` / `topK` (rejected by the newest models), and the temperature field is hidden for the Google provider in the options UI.
- **Schema**: the Google model schema is relaxed from a hardcoded enum to a non-empty string so dynamically discovered model ids validate; all other providers keep strict enum validation.
- Bumps `@ai-sdk/google` to `^4.0.21`.

## Related Issue

Closes #1938

## How Has This Been Tested?

- [x] Added unit tests

New tests cover the `fetchGoogleModels` fetcher (pagination, filtering, dedup, error surfacing), the model selector (dynamic options, API key kept out of the cache key), the relaxed Google schema, and the sampling-param-stripping middleware. Full suite: 215 test files / 2035 tests pass, `oxlint --type-aware` clean.

## Screenshots

See the reproduction screenshot in #1938.

## Checklist

## Additional Information

One follow-up worth discussing: existing users still have `reasoning: "none"` saved from the previous default. For the versionless aliases, `@ai-sdk/google` maps `"none"` to `thinkingConfig: { thinkingBudget: 0 }`, which the newest models reject — so those users would need to switch reasoning to "provider-default" manually if they pick an alias model. If you'd like, I can add a config migration (v86 → v87) that moves Google providers from `"none"` to `"provider-default"` in this PR or a follow-up.

### DIFF
--- a/.changeset/curvy-pandas-remember.md
+++ b/.changeset/curvy-pandas-remember.md
@@ -1,0 +1,5 @@
+---
+"@read-frog/extension": patch
+---
+
+fix(providers): load Gemini models dynamically
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@ai-sdk/deepinfra": "^3.0.12",
     "@ai-sdk/deepseek": "^3.0.12",
     "@ai-sdk/fireworks": "^3.0.13",
-    "@ai-sdk/google": "^4.0.18",
+    "@ai-sdk/google": "^4.0.21",
     "@ai-sdk/groq": "^4.0.12",
     "@ai-sdk/huggingface": "^2.0.12",
     "@ai-sdk/mistral": "^4.0.13",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -36,8 +36,8 @@ importers:
         specifier: ^3.0.13
         version: 3.0.13(zod@4.4.3)
       '@ai-sdk/google':
-        specifier: ^4.0.18
-        version: 4.0.18(zod@4.4.3)
+        specifier: ^4.0.21
+        version: 4.0.21(zod@4.4.3)
       '@ai-sdk/groq':
         specifier: ^4.0.12
         version: 4.0.12(zod@4.4.3)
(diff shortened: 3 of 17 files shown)


ananaBMaster:
should create pr to fix the hard coded model name. we won't support dynamic refresh only for Gemini. cuz we don't want the code to be full of special cases. not gonna have the time to maintain these special cases in the future

xuing:
> should create pr to fix the hard coded model name. we won't support dynamic refresh only for Gemini. cuz we don't want the code to be full of special cases. not gonna have the time to maintain these special cases in the future

ok, I understand.

PR #1944, opened 2026-07-23
feat(providers): auto-discover available models

xuing:
> **AI model(s) used (required):** Claude Fable 5, GPT-5.6 Sol Ultra 

## Type of Changes

- [x] ✨ New feature (feat)

## Description

The options page now quietly fetches each provider's current model list and merges it into the existing model dropdown, so newly released models are selectable without waiting for an extension release (the pain in #1938). No new UI and no user action needed.

**Design**

- One declarative registry (`MODEL_DISCOVERY_DESCRIPTORS`) maps a provider to `{ kind, baseURL }`; a `kind` is a shared response shape (`openai-compatible`, `google`, `anthropic`, `deepinfra`), not a per-provider code path. Adding/removing a provider is a one-line data change.
- Static `LLM_PROVIDER_MODELS` lists remain the source of truth and the always-available fallback; discovered ids only augment the dropdown. Picking an id outside the static enum persists through the existing `isCustomModel`/`customModel` channel — **no schema changes, no request-path changes**.
- Custom providers (OpenRouter, SiliconFlow, …) keep their static dropdown and join the same discovery path via their configured `baseURL`. Only `openai-compatible` keeps the manual custom-model input (its schema requires it), where the existing fetch button still works.
- Discovery reuses `getProviderHeadersWithOverride`, so default provider headers and user-configured `headers`/`baseURL` overrides behave exactly like translation requests.

**Performance / safety**

- Fetches only while the provider form is open, cached for 5 minutes per provider, in-flight requests cancelled via `AbortSignal` when the config changes, no retries.
- API key input is debounced and trimmed before any request; failures are silent and simply leave the static list in place (`meta.suppressToast`).
- Keys travel only in request headers — never in URLs; the react-query cache key uses a SHA-256 fingerprint, never the key itself.

Both Codex review points are addressed: DeepInfra discovery uses its public catalog (`/models/list`) instead of the chat-completions prefix, and named custom providers keep their static model dropdown.

## Related Issue

Refs #1938

## How Has This Been Tested?

- [x] Added unit tests

Unit tests cover the response shapes, pagination, capability filtering, header injection, baseURL overrides, and that keys never appear in URLs; component tests cover the automatic dropdown merge, custom-model routing, manual-entry toggle, debounce, and the silent-failure fallback. Full suite passes, `oxlint --type-aware` clean.

## Checklist

### DIFF
--- a/.changeset/eight-planets-listen.md
+++ b/.changeset/eight-planets-listen.md
@@ -1,0 +1,5 @@
+---
+"@read-frog/extension": minor
+---
+
+feat(providers): auto-discover available provider models
--- a/src/entrypoints/options/pages/api-providers/provider-config-form/__tests__/translate-model-selector.test.tsx
+++ b/src/entrypoints/options/pages/api-providers/provider-config-form/__tests__/translate-model-selector.test.tsx
@@ -317,28 +317,20 @@ describe("translateModelSelector", () => {
 
   it("loads public models automatically and routes a discovered id through custom model storage", async () => {
     const fetchMock = vi.fn<typeof fetch>().mockResolvedValue(
-      jsonResponse([
-        {
-          model_name: "new-chat-model",
-          reported_type: "text-generation",
-          deprecated: null,
-          private: 0,
-        },
-        {
-          model_name: "another-chat-model",
-          reported_type: "text-generation",
-          deprecated: null,
-          private: 0,
-        },
-        { model_name: "image-model", reported_type: "text-to-image" },
-      ]),
+      jsonResponse({
+        data: [
+          { id: "new-chat-model", metadata: { tags: ["chat"] } },
+          { id: "another-chat-model", metadata: { tags: ["chat"] } },
+          { id: "image-model", metadata: { tags: ["image-gen"] } },
+        ],
+      }),
     )
     vi.stubGlobal("fetch", fetchMock)
 
     renderSelector(deepInfraProviderConfig)
 
     await waitFor(() => {
-      expect(fetchMock).toHaveBeenCalledWith("https://api.deepinfra.com/models/list", {
+      expect(fetchMock).toHaveBeenCalledWith("https://api.deepinfra.com/v1/openai/models", {
         headers: {},
         signal: expect.any(AbortSignal),
       })
@@ -422,7 +414,7 @@ describe("translateModelSelector", () => {
     const fetchMock = vi
       .fn<typeof fetch>()
       .mockResolvedValue(
-        jsonResponse([{ model_name: "new-chat-model", reported_type: "text-generation" }]),
+        jsonResponse({ data: [{ id: "new-chat-model", metadata: { tags: ["chat"] } }] }),
       )
     vi.stubGlobal("fetch", fetchMock)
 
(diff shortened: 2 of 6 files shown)


mengxi-ream:
Thanks for putting this together and for the thorough test coverage. Automatic model discovery is useful, but we do not think the current persistence design is something we can merge.

The main blocker is that it removes a single source of truth for the selected model. When a discovered model is not in the bundled enum, the implementation keeps the previous enum-backed value in `model.model`, sets `isCustomModel: true`, and stores the actual selected model in `customModel`. This creates several problems:

- `model.model` can look authoritative while containing a stale fallback rather than the selected model.
- `isCustomModel` is redefined from manual custom entry to “anything not in the bundled static list,” even when the model came from the provider's official catalog.
- The UI presents the value as a normal discovered option, while the persisted configuration represents it as a custom model.
- The Zod schema validates the fields independently but does not express which value is authoritative or enforce the corresponding invariants.
- The representation can change across releases: a model discovered today is stored as custom even if it becomes a bundled model later.
- Every future consumer must remember to resolve the conditional fields correctly and must never read `model.model` directly.
- The selector also relies on overriding the autosave field's normal change handler to keep discovered IDs out of the enum field, which makes the design fragile and harder to maintain.

Avoiding a schema change moves the complexity into hidden state and ambiguous persisted semantics. We would prefer a single canonical model ID, with the source represented separately (for example, `builtin | discovered | manual`), or an equivalent discriminated-union design that makes the selected value unambiguous.

Because addressing this requires redesigning the underlying model representation rather than making another incremental patch, we are closing this PR. Thanks again for the contribution and for exploring this direction.

xuing:
> Thanks for putting this together and for the thorough test coverage. Automatic model discovery is useful, but we do not think the current persistence design is something we can merge.
> 
> The main blocker is that it removes a single source of truth for the selected model. When a discovered model is not in the bundled enum, the implementation keeps the previous enum-backed value in `model.model`, sets `isCustomModel: true`, and stores the actual selected model in `customModel`. This creates several problems:
> 
> * `model.model` can look authoritative while containing a stale fallback rather than the selected model.
> * `isCustomModel` is redefined from manual custom entry to “anything not in the bundled static list,” even when the model came from the provider's official catalog.
> * The UI presents the value as a normal discovered option, while the persisted configuration represents it as a custom model.
> * The Zod schema validates the fields independently but does not express which value is authoritative or enforce the corresponding invariants.
> * The representation can change across releases: a model discovered today is stored as custom even if it becomes a bundled model later.
> * Every future consumer must remember to resolve the conditional fields correctly and must never read `model.model` directly.
> * The selector also relies on overriding the autosave field's normal change handler to keep discovered IDs out of the enum field, which makes the design fragile and harder to maintain.
> 
> Avoiding a schema change moves the complexity into hidden state and ambiguous persisted semantics. We would prefer a single canonical model ID, with the source represented separately (for example, `builtin | discovered | manual`), or an equivalent discriminated-union design that makes the selected value unambiguous.
> 
> Because addressing this requires redesigning the underlying model representation rather than making another incremental patch, we are closing this PR. Thanks again for the contribution and for exploring this direction.

Fair. I understand. Thanks for your review.

PR #1981, opened 2026-07-28
fix(styles): apply appearance changes to live translations

JohnathanT55:
> **AI model(s) used (required):** OpenAI GPT-5 (Codex)

## Type of Changes

- [ ] ✨ New feature (feat)
- [x] 🐛 Bug fix (fix)
- [x] ✅ Test related (test)
- [ ] 🔄 CI/CD related (ci)
- [ ] 🧠 AI model related (ai)

## Description

Apply translation appearance changes to translations already present on an actively translated page.

The options preview cannot reproduce every website's typography and CSS. Refreshing the host page to verify an appearance change rebuilds the page state and translated DOM, which can be disruptive on long or interactive pages.

This implementation is lifecycle-scoped:

- registers the config listener only while page translation is active;
- disposes of the listener as soon as page translation stops;
- reacts only when `translationNodeStyle` changes;
- re-decorates existing block and inline translations, including translations inside open shadow roots;
- ensures translations that finish after a style change use the latest style;
- reconciles missed changes when a frozen tab becomes visible again;
- serializes style application to avoid concurrent custom stylesheet replacement;
- avoids another provider request or page-translation restart when only appearance changes.

## Related Issue

N/A

## How Has This Been Tested?

- [x] Added unit tests

Validation:

- full Vitest suite — 237 files and 2236 tests passed
- lifecycle/style targeted suite — 4 files and 10 tests passed
- type-aware lint and formatting checks passed for all changed files
- manually switched translation styles on an already translated page in Chrome and confirmed existing and late-arriving translations updated without another provider request

## Screenshots

N/A — this fixes live application of existing appearance settings rather than adding a new visual preset.

## Checklist

## Additional Information

- Includes a patch changeset for `@read-frog/extension`.
- No translation provider, prompt, or request scheduling behavior is changed.
- The branch is synchronized with the latest `main`, including #1982.


mengxi-ream (changes requested):
Hi,

We are reluctant to add the listener for this, since users can refresh the page to get the new translation style.

Any specific reason you think we should merge this PR?

JohnathanT55:
> Hi,
> 
> We are reluctant to add the listener for this, since users can refresh the page to get the new translation style.
> 
> Any specific reason you think we should merge this PR?


Hi, thanks for the feedback.

The main use case is adjusting translation appearance while actively reading. The preview in the options page cannot fully reproduce a website's own typography and CSS, so users may need to verify colors, font size, and other appearance settings on the actual translated page.

Refreshing rebuilds both the page state and the translated DOM, which can be disruptive on long or interactive pages. Applying the new style directly to existing translation nodes provides immediate feedback without restarting the page translation flow.

This also seems consistent with the recent same-origin navigation change in #1982, which preserves the live translation session instead of tearing it down and rebuilding it, avoiding translated-content flicker.

I understand the concern about adding a broad config listener to every host content script. I can narrow the implementation so that the listener is registered only while page translation is active, disposed of as soon as translation stops, and reacts only when `translationNodeStyle` changes.

Would that narrower lifecycle-scoped implementation address your concern?

mengxi-ream:
Hi, thank you for the update! Could you give me an example website you think it really need instant change of the translation style css?

PR #1996, opened 2026-07-30
fix: translation hub data persistence

### LINKED ISSUE #1633: [FEATURE] 翻译中心记住提供商/模型设置 translation hub
### Before you begin

- [x] I am willing to contribute a PR for this feature

### Is your feature request related to a problem? Please describe.

每次刷新后勾选的“翻译服务”设置就被重置了，非常不方便。

### Describe the solution you'd like

希望能记住配置好的选项。

### Which area(s) would this feature affect? (Select all that apply)

Browser Extension, Website

### Additional context

相关issue #925，但提出者否认了这个角度。

li-yiou:
> **AI model(s) used (required):** 

## Type of Changes

- [ ] ✨ New feature (feat)
- [x] 🐛 Bug fix (fix)
- [ ] ✅ Test related (test)
- [ ] 🔄 CI/CD related (ci)
- [ ] 🧠 AI model related (ai)

## Description

Fixed data persistence of translation hub.

## Related Issue

Closes #1633 #1643 

## How Has This Been Tested?

- [ ] Added unit tests

## Checklist

### DIFF
--- a/src/entrypoints/translation-hub/atoms.ts
+++ b/src/entrypoints/translation-hub/atoms.ts
@@ -4,26 +4,16 @@ import { atom } from "jotai"
 import { configFieldsAtomMap } from "@/utils/atoms/config"
 import { filterEnabledProvidersConfig, getTranslateProvidersConfig } from "@/utils/config/helpers"
 
-// === LangCode Atoms (derive from config, local override) ===
-const sourceLangCodeOverrideAtom = atom<LangCodeISO6393 | "auto" | null>(null)
-const targetLangCodeOverrideAtom = atom<LangCodeISO6393 | null>(null)
-
+// === LangCode Atoms (persist through the shared language config) ===
 export const sourceLangCodeAtom = atom(
-  (get) => {
-    const override = get(sourceLangCodeOverrideAtom)
-    if (override !== null) return override
-    return get(configFieldsAtomMap.language).sourceCode
-  },
-  (_get, set, value: LangCodeISO6393 | "auto") => set(sourceLangCodeOverrideAtom, value),
+  (get) => get(configFieldsAtomMap.language).sourceCode,
+  (_get, set, value: LangCodeISO6393 | "auto") =>
+    set(configFieldsAtomMap.language, { sourceCode: value }),
 )
 
 export const targetLangCodeAtom = atom(
-  (get) => {
-    const override = get(targetLangCodeOverrideAtom)
-    if (override !== null) return override
-    return get(configFieldsAtomMap.language).targetCode
-  },
-  (_get, set, value: LangCodeISO6393) => set(targetLangCodeOverrideAtom, value),
+  (get) => get(configFieldsAtomMap.language).targetCode,
+  (_get, set, value: LangCodeISO6393) => set(configFieldsAtomMap.language, { targetCode: value }),
 )
 
 // === Input Atom ===
@@ -32,19 +22,21 @@ export const inputTextAtom = atom("")
 // === Detected Source LangCode (from input text) ===
 export const detectedSourceLangCodeAtom = atom<LangCodeISO6393 | null>(null)
 
-// === Selected Provider IDs (only store IDs, get config from configFieldsAtomMap) ===
-const selectedProviderIdsOverrideAtom = atom<string[] | null>(null)
-
+// === Selected Provider IDs (persist IDs and derive live provider config) ===
 export const selectedProviderIdsAtom = atom(
   (get) => {
-    const override = get(selectedProviderIdsOverrideAtom)
-    if (override !== null) return override
-    // Default: all enabled translate providers' IDs
     const providersConfig = get(configFieldsAtomMap.providersConfig)
     const translateProviders = getTranslateProvidersConfig(providersConfig)
-    return filterEnabledProvidersConfig(translateProviders).map((p) => p.id)
+    const enabledIds = filterEnabledProvidersConfig(translateProviders).map((p) => p.id)
+    const selectedIds = get(configFieldsAtomMap.translationHub).selectedProviderIds
+
+    if (selectedIds === null) return enabledIds
+
+    const enabledIdSet = new Set(enabledIds)
+    return selectedIds.filter((id) => enabledIdSet.has(id))
   },
-  (_get, set, ids: string[]) => set(selectedProviderIdsOverrideAtom, ids),
+  (_get, set, ids: string[]) =>
+    set(configFieldsAtomMap.translationHub, { selectedProviderIds: ids }),
 )
 
 // === Translation Card UI State ===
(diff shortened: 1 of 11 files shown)


mengxi-ream (dismissed):
Superseded — see the line-by-line review below.

src/entrypoints/translation-hub/atoms.ts
```diff
@@ -4,26 +4,16 @@ import { atom } from "jotai"
...
 export const sourceLangCodeAtom = atom(
-  (get) => {
-    const override = get(sourceLangCodeOverrideAtom)
-    if (override !== null) return override
-    return get(configFieldsAtomMap.language).sourceCode
-  },
-  (_get, set, value: LangCodeISO6393 | "auto") => set(sourceLangCodeOverrideAtom, value),
+  (get) => get(configFieldsAtomMap.language).sourceCode,
+  (_get, set, value: LangCodeISO6393 | "auto") =>
+    set(configFieldsAtomMap.language, { sourceCode: value }),
```
mengxi-ream, line 11: **Blocking.** `config.language` is the extension-wide page-translation language — the same field the popup's main selector writes (`popup/components/language-options-selector.tsx:99`). Writing it from the hub means it gets read back by:

- `utils/host/translate/translate-variants.ts:109` (page translation core)
- `utils/host/translate/target-language-skip.ts:21`
- `utils/host/translate/filter-small-paragraph.ts:27`
- `utils/subtitles/processor/translator.ts:211`
- `selection.content/selection-toolbar/atoms.ts:105`
- `background/analytics.ts:220`

So a user who opens the Translation Hub to translate one Chinese snippet into English has silently retargeted every webpage translation and every subtitle track to English.

#1643 asks to remember the direction **for the hub**, not to share the global one. Please put `sourceLangCode` / `targetLangCode` in the `translationHub` section this PR already introduces — same amount of code, symmetric with `selectedProviderIds`, and no cross-feature side effects. The provider half of this PR is done exactly right; please do the language half the same way.

src/entrypoints/translation-hub/components/language-control-panel.tsx
```diff
@@ -58,7 +58,7 @@ export function LanguageControlPanel() {
         <Button
           variant="ghost"
           size="icon"
-          onClick={exchangeLangCodes}
+          onClick={() => void exchangeLangCodes()}
```
mengxi-ream, line 61: Follow-on from the atoms change: this button now swaps the global `config.language`, so one click here flips the translation direction for every webpage and every subtitle track, not just this panel. Goes away once the language fields move into `translationHub`.

src/utils/config/__tests__/example/v089.ts
```diff
@@ -0,0 +1,50 @@
+import type { TestSeriesObject } from "./types"
+import { testSeries as previousTestSeries } from "./v088"
+
+const complexConfig = previousTestSeries["complex-config-from-v020"]!
+const llmDetectionConfig = previousTestSeries["config-with-llm-detection-enabled"]!
+const promptTokenConfig = previousTestSeries["prompt-token-migration-coverage"]!
+const defaultDictionaryWordingConfig = previousTestSeries["default-dictionary-wording"]!
```
mengxi-ream, line 7: **Blocking.** The doc comment just below says "Frozen v089 expectations", but this spreads from v088 — and v088 spreads from v087, so it's a two-level chain now.

`all-migrations.test.ts:66-68` asserts `migrate(vN fixture)` equals the `vN+1` fixture. That assertion only means anything when both sides are independent literals. With spreads, editing v087 shifts the v088 **and** v089 expectations together and the test stays green — the regression net is gone.

Symptom that's already visible: `grep "description:" v089.ts` returns nothing. Both v088 and v089 inherit v087's `"Adds opacity to selection toolbar"`, which describes neither change.

Please expand this into a full standalone literal, keeping only the type-only import on line 1, the way v087 and everything before it are. ~1700 lines is normal and expected for these fixtures.

To be fair: v088 (#1980) is where this pattern actually broke, not your PR — you followed the most recent example. Backfilling v088 is on us as a separate PR; you only need to fix v089 here.

src/types/config/config.ts
```diff
@@ -85,6 +85,12 @@ const contextMenuSchema = z.object({
   enabled: z.boolean(),
 })
 
+const translationHubSchema = z
+  .object({
+    selectedProviderIds: z.array(z.string().nonempty()).nullable(),
```
mengxi-ream, line 90: Worth a comment that `null` and `[]` mean different things here: `null` = never chosen, follow all enabled providers; `[]` = explicitly chose none. Your test encodes this (`atoms.test.ts:49-59`) but the schema doesn't say it anywhere.

For what it's worth, I did look for a simpler encoding than the tri-state and there isn't one — you need to distinguish "new user, show everything" from "explicit empty selection sticks". The complexity is justified, it just needs labelling.

src/types/config/config.ts
```diff
@@ -85,6 +85,12 @@ const contextMenuSchema = z.object({
...
+  })
+  .default({ selectedProviderIds: null })
```
mengxi-ream, line 92: This `.default()` is correct — please **don't** remove it — but document why, the way `uiLanguageSchema` does on lines 118-123.

`storage-adapter.ts:11` silently returns `fallback` (i.e. `DEFAULT_CONFIG`) when the schema parse fails. `initializeConfig` only runs in the background script, so a content script in an already-open tab can load the new code and try to parse v088-shaped storage against the v089 schema before the migration lands. Without this `.default()`, that parse fails, the atom gets `DEFAULT_CONFIG`, and the next `writeConfigAtom` overwrites the user's entire config.

With no comment here, a future cleanup will read it as "the migration already adds this field, the default is redundant" and delete it.

src/entrypoints/translation-hub/atoms.ts
```diff
@@ -32,19 +22,21 @@ export const inputTextAtom = atom("")
...
     const providersConfig = get(configFieldsAtomMap.providersConfig)
     const translateProviders = getTranslateProvidersConfig(providersConfig)
-    return filterEnabledProvidersConfig(translateProviders).map((p) => p.id)
+    const enabledIds = filterEnabledProvidersConfig(translateProviders).map((p) => p.id)
+    const selectedIds = get(configFieldsAtomMap.translationHub).selectedProviderIds
+
+    if (selectedIds === null) return enabledIds
+
+    const enabledIdSet = new Set(enabledIds)
+    return selectedIds.filter((id) => enabledIdSet.has(id))
```
mengxi-ream, line 36: Nit: this filter drops IDs of providers that are currently disabled, and `translation-service-dropdown.tsx:51` feeds the filtered list straight back through `onValueChange`. So: select [A, B] → disable B in settings → hub shows [A] → touch the dropdown → B is persisted away for good, and re-enabling B later won't bring it back.

If you want a selection to survive a temporary disable, merge the write against the stored list rather than the filtered one.

src/entrypoints/translation-hub/atoms.ts
```diff
@@ -60,12 +52,14 @@ export const selectedProvidersAtom = atom((get) => {
...
-  set(sourceLangCodeAtom, target)
-  set(targetLangCodeAtom, source)
+export const exchangeLangCodesAtom = atom(null, async (get, set) => {
+  const language = get(configFieldsAtomMap.language)
+  if (language.sourceCode === "auto") return // Cannot exchange when source is auto
+
+  await set(configFieldsAtomMap.language, {
+    sourceCode: language.targetCode,
+    targetCode: language.sourceCode,
+  })
```
mengxi-ream, line 62: Collapsing the two sequential writes into one is a real improvement — the old version raced two `writeConfigAtom` calls. Keep that part.

The `async` / `await` isn't doing anything though: nothing awaits this (`language-control-panel.tsx:61` is `void exchangeLangCodes()`), and `writeConfigAtom` already does its optimistic atom update synchronously before queueing the storage write.

mengxi-ream (changes requested):
Thanks for this. `tsc`, `oxlint` and the full test suite all pass, and the migration script itself is textbook — hardcoded values, `any` in/out, no `src/` imports, idempotent and covered by a test. Auto-registration through `import.meta.glob` works, v089 is still a free slot on `main`, and the `void` wrappers match how the rest of the codebase calls config setters.

Two blocking items, both commented inline: the language atoms write the extension-wide `config.language` instead of the new `translationHub` section, and `example/v089.ts` isn't a standalone fixture. The rest are comments and nits.

#1633 is fully addressed. #1643 also asked for a settings-page default for the hub's language direction, which isn't here — fine to defer, just don't auto-close that one on merge.

PR #2014, opened 2026-07-31
fix(translate): translate SillyTavern <code> narration via site-rule tag-set overrides (#1951)

### LINKED ISSUE #1951: [BUG]只翻译对话的bug？
### Before you report

不知为何只翻译“对话”的文本，其他地方依旧是原文，已经尝试过切换翻译模型/清除数据/重启重装游览器/重装插件，依旧无效。使用“”和‘’之间的文字会被翻译，但其他部分无法翻译（似乎还有文本量的问题，小量的话也可以全部翻译），但是‘’的话全文都可以被翻译
错误情况大概如下：
Image

### To Reproduce

不知为何只翻译“对话”的文本，其他地方依旧是原文，已经尝试过切换翻译模型/清除数据/重启重装游览器/重装插件，依旧无效。使用“”和‘’之间的文字会被翻译，但其他部分无法翻译（似乎还有文本量的问题，小量的话也可以全部翻译），但是‘’的话全文都可以被翻译
错误情况大概如下：
Image

Browser Extension

1.42.3

```bash
- OS:Android 16
- Browser:Edge
- Extension channel/version: 1.42.3
- Website: sillytavern，07.23
```

### Additional context

目前看起来只有sillytavern有这个情况，但我确实之前都是没问题的。像是纯文本翻译都没问题。

JoeJoeflyn:
Fixes #1951.

## Root cause

SillyTavern renders backtick narration as `<p><code>…</code></p>`. `CODE` is in `DONT_WALK_BUT_TRANSLATE_TAGS`, so a code-only paragraph never sets `hasInlineNodeChild`, never gets `PARAGRAPH_ATTRIBUTE`, and is skipped — while `extractTextContent` already includes code text whenever any sibling text exists (the labeling layer contradicted the category's own contract).

## Approach (supersedes the original `translateAsChildSelectors` proposal)

Instead of special-casing the walk, the five hardcoded DOM tag-set constants become per-site overridable through the existing site-rules delta machinery:

- New rule fields `dontWalkTags` / `dontWalkButTranslateTags` / `mainContentIgnoreTags` / `forceBlockTags` / `forceInlineTranslationTags`, each **`.add`/`.remove` only** (delta-only, no bare key) with **tag-name** entries — resolution materializes a `Set`, so the per-element hot path stays `Set.has` (#1881 untouched). Defaults are seeded in code inside `resolveSiteRule`; `null` means "family untouched → constant fallback".
- Removal of catastrophic tags (`HEAD/TITLE/META/SCRIPT/NOSCRIPT/STYLE/LINK`) is refused with a warning; `PRE` stays removable by design. `notranslate`/hidden/preserveText checks remain independent and un-overridable.
- Built-in rule `sillytavern` (`matches: ["localhost", "127.0.0.1"]`, `"dontWalkButTranslateTags.remove": ["CODE"]`): `<code>` becomes a normally walked inline element there and code-only paragraphs translate exactly like their `<q>` dialogue siblings — no changes to `traversal.ts`/`find.ts` needed. Fenced code blocks stay protected by `PRE`. LAN-IP/reverse-proxy users can add a user rule with the same field (docs follow-up).

## Commits

1. `feat(site-rules)` — schema keys + `mergeTagSetDelta` resolver support
2. `refactor(dom)` — thread all five constants' consumers through `getEffectiveTagSet` (byte-identical behavior with no matching rule)
3. `fix(translate)` — the sillytavern built-in rule + end-to-end tests + changeset
4. `test` — hardening from adversarial review: pin the jsdom test URL to a neutral host (jsdom's `localhost:3000` default silently matched the new rule), consumer-level tests for the two merge-only families, and an unknown-key check for `rules.json`

## Testing

- Full suite: 2465 passed; `pnpm type-check` clean.
- New coverage: resolver merge semantics (seeding, ordering, case tri-form, protected tags), effective-rule memo Set identity, editor schema accept/reject, `localhost` pattern matching, filter predicate overrides, and end-to-end walk/extract tests for the SillyTavern markup (labeled on localhost, baseline preserved elsewhere, `<pre><code>` still blocked, rule disable restores default).

Credit to @JoeJoeflyn for reproducing #1951 and pinning down the root cause.

### DIFF
--- a/.changeset/site-rule-tag-set-overrides.md
+++ b/.changeset/site-rule-tag-set-overrides.md
@@ -1,0 +1,5 @@
+---
+"@read-frog/extension": patch
+---
+
+feat(site-rules): allow per-site `.add`/`.remove` overrides of the built-in DOM tag sets (dontWalkTags, dontWalkButTranslateTags, mainContentIgnoreTags, forceBlockTags, forceInlineTranslationTags), and ship a localhost rule that lets SillyTavern's `<p><code>` narration translate (#1951)
--- a/src/entrypoints/options/pages/translation/translation-control/site-rules/__tests__/validate-user-rules.test.ts
+++ b/src/entrypoints/options/pages/translation/translation-control/site-rules/__tests__/validate-user-rules.test.ts
@@ -123,6 +123,34 @@ describe("validateUserRulesDocument", () => {
     ])
   })
 
+  it("accepts the tag-set delta fields", () => {
+    const rules = [
+      {
+        id: "tag-sets",
+        matches: "example.com",
+        "dontWalkButTranslateTags.remove": ["CODE"],
+        "dontWalkTags.add": ["X-WIDGET"],
+      },
+    ]
+
+    expect(validateUserRulesDocument(JSON.stringify(rules))).toEqual({ ok: true, rules })
+  })
+
+  it("rejects a bare tag-set key to keep the fields delta-only", () => {
+    const result = validateUserRulesDocument(
+      JSON.stringify([{ id: "bare", matches: "example.com", dontWalkTags: ["CODE"] }]),
+    )
+
+    expect(result.ok).toBe(false)
+    if (result.ok) {
+      return
+    }
+    expect(result.kind).toBe("schema")
+    expect(result.issues).toEqual([
+      { path: "rules[0].dontWalkTags", message: 'Unrecognized field "dontWalkTags"' },
+    ])
+  })
+
   it("reports invalid legacy selector values at their legacy field paths", () => {
     const result = validateUserRulesDocument(
       JSON.stringify([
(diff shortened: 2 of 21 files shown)


mengxi-ream (changes requested):
Hi,

based on your video (https://github.com/mengxi-ream/read-frog/issues/1951#issuecomment-5151257282), I think the root cause is the original website put the text into the `<code>` but it's normal for all other website to not translate `<code>`, since ... it's code.

I think your fix may translate the `<code>` for all other website which people don't want to. So what you need to do is to add the special rule in our rule system rather than using the method you currently use in this PR.

Also another reminder is that, Coding AI from Google and GLM is not good enough and we suggest to use Codex, Claude or at lease Kimi K3

mengxi-ream:
But your reproduction video is super helpful. I really appreciate your time and effort!

JoeJoeflyn:
Thanks! Fair point, I'll do it as a site rule instead. Noted on the model too.

mengxi-ream:
also you can include the demo to show "before" "after" effects

JoeJoeflyn:
Fixed!

mengxi-ream:
> Fixed!

did you used our rule system? I don't see the code about it

JoeJoeflyn:
My bad, the merge with main dropped the fix. Just pushed it again with the site rule version

PR #2018, opened 2026-07-31
feat(subtitles): add a configurable shortcut to toggle subtitle trans…

### LINKED ISSUE #1979: [FEATURE] Add a keyboard shortcut to toggle bilingual subtitles (or sync with YouTube's `C` key)
### Is your feature request related to a problem?

When watching YouTube with bilingual subtitles enabled by default, I often want to temporarily hide them and only view the original subtitles.

Currently, the only way to do this is to open the Read Frog UI and click the toggle every time. This interrupts the viewing experience.

### Proposed solution

Please consider one (or both) of the following:

1. Add a dedicated keyboard shortcut to toggle Read Frog bilingual subtitles on/off.
2. Optionally allow the toggle to follow YouTube's native subtitle shortcut (`C`), so enabling/disabling subtitles also enables/disables Read Frog subtitles.

### Why this is useful

- Quickly compare original subtitles with translated subtitles.
- Less mouse interaction while watching videos.
- Better for language learning, where users frequently switch between bilingual and original subtitles.

### Additional suggestion

The best long-term solution may be exposing this as a configurable extension command so users can bind their own shortcut from the browser's extension shortcut settings.

ananaBMaster:
> **AI model(s) used (required):** Claude Opus 5

## Type of Changes

- [x] ✨ New feature (feat)

## Description

Watching a video with bilingual subtitles enabled, the only way to temporarily fall back to the original captions was to open the Read Frog panel and click the ON/OFF switch every time — which interrupts playback, and is especially tedious for language learners who flip back and forth constantly.

This adds a configurable keyboard shortcut that does exactly what that switch does.

**Behavior.** Pressing the shortcut flips the same runtime visibility the panel switch controls: turning it off restores the site's own captions and stops the translation coordinator; turning it on brings the overlay back. A small toast confirms each flip. It is deliberately *not* a `style.displayMode` change — translation actually stops, rather than continuing to burn requests behind a hidden line.

**Default key: `Alt+C`.** Mnemonic for "captions", and it echoes the native `C` key without taking it over. It does not collide with the extension's own `Alt+E` / `Alt+T` / `Alt+Shift+M` defaults, with the video player's own shortcuts (all bare keys), with Chrome, or with Firefox's Alt menu mnemonics (F/E/V/S/B/T/H). On macOS Option+C types `ç`, which `@tanstack/hotkeys` handles via its `event.code` fallback — verified in a real browser, not just assumed.

**Migration.** `v088-to-v089` adds `videoSubtitles.toggleShortcut`. If the user has already bound `Alt+C` to page translation, translation mode, or selection translation, it falls back to `Alt+Shift+C`; if both are taken, it ships unbound. That fallback key exists **only** inside the frozen migration script, never in shared constants — it is meaningful only for configs that predate this field.

**Where the binding lives.** In `subtitles.content`, because that is where the `UniversalVideoAdapter` lives and there is no `ProtocolMap` message for toggling subtitle visibility. It takes the shared `SubtitlesProvidersAdapter` interface rather than the YouTube adapter, so it is not platform-specific: a future platform reuses it by calling the same binder from its own init. The WXT `ctx` is now threaded through `runtime.ts` → `init-youtube-subtitles.ts` so the hotkey registration unbinds on invalidation.

### Files of note

| File | What |
|---|---|
| `src/entrypoints/subtitles.content/bind-subtitles-toggle-shortcut.ts` | New binding, mirroring `bind-translation-mode-shortcut.ts` |
| `src/utils/config/migration-scripts/v088-to-v089.ts` | New field + collision-aware fallback |
| `src/entrypoints/options/pages/shortcuts/subtitles-toggle-shortcut.tsx` | New row on the Shortcuts page |
| `src/types/config/subtitles.ts` | `toggleShortcut`, reusing `pageTranslationShortcutSchema` |
| `src/entrypoints/subtitles.content/universal-adapter.ts` | `toggleSubtitlesByShortcut`; toggle source → analytics surface lookup |

Also registered in `SEARCH_ITEMS` for the command palette, linked from the Video Subtitles page via `ShortcutLink`, and translated into all 9 locales.

Following the player's native `C` key is intentionally out of scope — it needs a caption-state observer plus careful focus handling, and deserves its own issue.

## Related Issue

Closes #1979

## How Has This Been Tested?

- [x] Added unit tests

**Unit tests** — `v088-to-v089.test.ts` covers default assignment, the fallback for each of the three colliding shortcut fields, case/whitespace-insensitive collision detection, the both-taken → unbound case, idempotency, and malformed input. The Shortcuts page ordering test was updated. Full suite: 2323 passed, `pnpm lint` and `pnpm fmt:check` clean.

**Real-browser E2E** — drove the built extension in Chrome against a live video page via Puppeteer, 8/8:

1. Default ships as `Alt+C`
2. Shortcuts page renders the new row; the recorder displays `⌥ C`, matching the existing `⌥ E` / `⌥ ⇧ M` / `⌥ T` rows
3. Rebinding through the recorder persists `Alt+Shift+C` to storage
4. The subtitles content script mounts its player button
5. `Alt+C` turns Read Frog subtitles on
6. `Alt+C` again turns them back off
7. Pressing it while focused in the site's search box does nothing (`ignoreInputs`)
8. No page errors

## Additional Information

Two things worth a reviewer's attention:

1. **Rebinding takes effect on the next page load.** The key is read once at bind time. This matches every other shortcut in the extension (`bind-translation-shortcut.ts`, `bind-translation-mode-shortcut.ts`), so it is consistent rather than a new wart — but it is a real limitation if we ever want live rebinding.
2. **`example/v089.ts` is written out in full** (~47KB, a standalone literal with only a type-only import), matching v070–v087. A fixture that spreads the previous version is not a frozen snapshot: editing an older fixture would silently change what the newer one asserts. Note that `example/v088.ts` currently *does* spread from v087 — pre-existing, left alone here to keep this diff focused, but worth fixing separately.


ananaBMaster:
@taiiiyang could you review this PR?

PR #2030, opened 2026-08-02
fix(translate): correct X tweet formatting — note-tweet paragraph interleaving + Google line-break preservation

mengxi-ream:
## Problem

Two independent bugs broke translation formatting on X/Twitter.

Reported cases:
- https://x.com/davidjpark96/status/1789773192435060737 — 22k-char note tweet: translations displaced/merged (bug 1), and its bullet list squashed onto one line under Google Translate (bug 2)
- https://x.com/EpsteinJeffrey0/status/2083709421386080579 — regular tweet of five single-`\n` lines: Google merged them into one run-on translation (bug 2)

**1. Long note tweets: translations displaced / merged (any provider)**

A 20k+-char note tweet renders as ONE pre-wrap `tweetText` div whose rich-text runs are sibling inline `<span>`s — paragraphs are literal blank lines *inside* the spans, so span boundaries ≠ paragraph boundaries. The #1881 giant-paragraph split (>3 viewports) observed each span as an independent translation unit, and an inline span can never enter the virtual-paragraph path. Every span translated as one blob whose wrapper landed at the span's end:

- bold heading spans got inline translations glued mid-line (`- 付费广告Organic Short-form content  有机短视频内容`)
- a 5k-char plain span's translation landed ~25 paragraphs below its content

**2. Google Translate: line breaks eaten (any multi-line tweet)**

Google's `translateHtml` endpoint parses plain text as HTML, so literal `\n` collapses as whitespace — bullet lists and single-newline tweets became one merged run-on line with merged-sentence translations. No escape survives (`&#10;` collapses too).

## Fix

**Commit 1 — `canSplitParagraphIntoDescendants` guard:** a newline-preserving container whose descendant paragraphs are all inline is never split — observed whole, it routes through the virtual-paragraph plan, which interleaves translations per blank-line paragraph. Containers with block descendants (chat-style pre-wrap lists) keep splitting, preserving #1881 viewport gating. The predicate uses the same effective block marker the walker dispatches on.

**Commits 2+4 — `preserveLineBreaks` flag + whole-unit marker transport:** the caller signals that line breaks are semantic (newline-preserving flow container / typed input); only the Google provider acts on it, per live-verified endpoint behavior:

- lines are joined into **ONE request item with `<br>` marker pairs** (a lone `<br>` can be merged away by the sentence segmenter; a pair never is) — so the endpoint language-detects the **whole unit** and `sl=auto` is never overridden. This is the correctness invariant: per-line items are detected independently and short lines misread (`- SEO` alone → `- 这`), while inside its own tweet the same line translates correctly. A Japanese tweet on an English page also stays correct — the extension never forces a language (E2E-verified).
- the model strips leading dash bullets from list lines regardless of transport → indentation + bullet prefix captured client-side and restored per marker segment; graceful marker-fold fallback if segment counts ever drift
- the giant-split refusal from commit 1 applies only in **bilingual** mode (translationOnly swaps text in place per span and keeps viewport gating); the flag's gate samples the **flow container's** white-space (a GitHub inline `<code>` with `break-spaces` no longer flags a normal paragraph)

An earlier iteration (per-line batch items + page-level detected-language override) was found unsound by adversarial review — the override broke mixed-language pages and the detection default (`und` → `eng`) forced English during the startup race — and was replaced wholesale; the `detectedSourceCode` threading is deleted. Cache hash gains a `preserveLineBreaks` component only for Google requests, keeping existing entries and other providers' entries valid. Microsoft is unchanged (it already preserves `\n`); LLM prompts already mandate format preservation. Known gap: translationOnly's html-format payloads still collapse (serialization-layer work, documented in `google.ts`).

## Verification

- Puppeteer E2E against the built extension on faithful fixtures of both reported tweets (22k-char rich-text note with bold heading spans; single-`\n` 5-line tweet): before = pixel-level reproduction of the reported layouts; after = 141 correctly interleaved per-paragraph wrappers, bullets/dashes/lines all preserved, with both Microsoft and Google
- Both reported URLs are referenced in code comments at the fix sites (`paragraph-segmentation.ts`, `page-translation.ts`, `google.ts`, `translation-modes.ts`) and in the regression tests (`paragraph-segmentation.test.ts` note-tweet split case; `google-roundtrip.test.ts` bullet-list and plain single-newline cases)
- All endpoint behaviors (newline collapse, dash eating, per-item auto misdetection, multi-item form, empty-item 400) verified against the live `translateHtml` API
- Full suite: 2455 tests pass; new regression tests for the split guard and a Google round-trip simulator emulating collapse/dash-eating/indent-stripping
- Multi-agent adversarial review of commit 1 confirmed + fixed one guard-predicate edge (mixed block+inline giants)

### DIFF
--- a/.changeset/x-tweet-line-structure.md
+++ b/.changeset/x-tweet-line-structure.md
@@ -1,0 +1,5 @@
+---
+"@read-frog/extension": patch
+---
+
+fix(translate): correct X tweet formatting — interleave long note-tweet paragraphs with their translations, and preserve line breaks and list dashes through Google Translate
--- a/src/entrypoints/background/translation-queues.ts
+++ b/src/entrypoints/background/translation-queues.ts
@@ -470,7 +470,6 @@ export function setUpWebPageTranslationQueue(): void {
         hash,
         textFormat,
         preserveLineBreaks,
-        detectedSourceCode,
         webTitle,
         webDescription,
         webContent,
@@ -566,7 +565,6 @@ export function setUpWebPageTranslationQueue(): void {
         executeTranslate(text, langConfig, providerConfig, getTranslatePrompt, {
           textFormat,
           preserveLineBreaks,
-          detectedSourceCode,
           signal,
         })
       result = await requestQueue.enqueue(thunk, scheduleAt, hash, scope ? [scope] : undefined)
--- a/src/entrypoints/background/translation-queues.ts
+++ b/src/entrypoints/background/translation-queues.ts
@@ -469,6 +469,8 @@ export function setUpWebPageTranslationQueue(): void {
         scheduleAt,
         hash,
         textFormat,
+        preserveLineBreaks,
+        detectedSourceCode,
         webTitle,
         webDescription,
         webContent,
@@ -563,6 +565,8 @@ export function setUpWebPageTranslationQueue(): void {
       const thunk = (signal?: AbortSignal) =>
         executeTranslate(text, langConfig, providerConfig, getTranslatePrompt, {
           textFormat,
+          preserveLineBreaks,
+          detectedSourceCode,
           signal,
         })
       result = await requestQueue.enqueue(thunk, scheduleAt, hash, scope ? [scope] : undefined)
--- a/src/entrypoints/host.content/translation-control/page-translation.ts
+++ b/src/entrypoints/host.content/translation-control/page-translation.ts
@@ -44,6 +44,7 @@ import {
   wasCharacterDataChangeExtensionDriven,
   wasNodeRemovedByExtension,
 } from "@/utils/host/translate/core/translation-state"
+import { canSplitParagraphIntoDescendants } from "@/utils/host/translate/dom/paragraph-segmentation"
 import {
   removeAllTranslatedWrapperNodes,
   translateNodes,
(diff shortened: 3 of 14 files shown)


mengxi-ream:
Re the P1 on the page-detected `sl` override (`translate-variants.ts`): confirmed and fixed in 0560c7c63 — by deletion rather than repair. An independent multi-agent adversarial review converged on the same defect (plus the `und`→`eng` normalization race, which made the `?? undefined` dead code).

The override existed to patch per-item auto-detection misreading short lines (`- SEO` alone → `- 这`, live-verified). The rework removes the need for it: lines now travel as **one request item joined by `<br>` marker pairs**, so the endpoint detects the language of the whole unit itself — `- SEO` translates correctly inside its own tweet, a Japanese tweet on an English-detected page stays correct (E2E-verified), and `sl=auto` is never overridden. The entire `detectedSourceCode` threading is gone.

Also folded in from the same review: giant-split refusal gated to bilingual mode, flow-container white-space sampling (GitHub inline `code` false positive), bullet-only-line dash doubling, and the cache-hash component scoped to Google.

mengxi-ream:
Re the P1 on the split refusal for flows without blank-line delimiters: correct, fixed in 99ac88d25. `canSplitParagraphIntoDescendants` now consults the same plan builder the translate path uses and refuses the split only when the container actually yields ≥ 2 virtual paragraphs — a single-newline-only giant (pre-wrapped log/code views) keeps the descendant split and its viewport gating. Regression test added for that shape.

PR #2072, opened 2026-08-14
feat(subtitles): take AI subtitles out of beta behind a Pro/Ultra minute quota

mengxi-ream:
> **AI model(s) used (required):** Claude Fable 5 (claude-fable-5)

## Type of Changes

- [x] ✨ New feature (feat)

## Description

Client half of the AI-subtitles GA launch (server half: mengxi-ream/read-frog-monorepo — "take AI subtitles out of beta behind a Pro/Ultra minute quota"). AI transcription leaves beta and becomes a Pro/Ultra feature with a minute quota, so the extension stops pre-flighting beta access and starts rendering the subscription wall and the per-pool quota the server now reports.

**Request path**

- `access-guard.ts`: the beta pre-flight and the Tally application form are gone. Sign-in is the only client-side pre-check — the subscription wall is the server's `create` response, so there is no second source of truth to drift.
- `create` now sends `durationSec`, read from the player's media element (`Math.ceil(video.duration)`). It is an untrusted admission hint only — the server re-measures with ffprobe in its worker before billing — so the ad-playback edge (where `duration` briefly reports the ad's length) is harmless by design.
- New error-code branches in `request-ai-subtitles.ts`:
  - `VIDEO_TRANSCRIPTION_SUBSCRIPTION_REQUIRED` → toast + open `/pricing` (via background `openPage`; content scripts can't use `chrome.tabs`)
  - `VIDEO_TRANSCRIPTION_PAYMENT_REQUIRED` → toast + open `/home` (dunning, not cancellation — an existing subscriber shouldn't land on pricing)
  - `VIDEO_TRANSCRIPTION_UNSUPPORTED_LENGTH` → toast only (a property of the video, not the account — neither upgrading nor waiting fixes it)

**Options page quota section**

- One progress bar per quota pool: the monthly subscription pool labeled with its reset date, and — for launch-window subscribers — the one-time launch gift labeled with its expiry date. Near-limit (≥90%) still turns the bar destructive.
- A pre-pools server's totals-only response renders as the single bar it always was; a pre-launch server's beta 403 and the new server's `plan: "free"` both render the same upgrade card.
- Usage is fetched only when the section mounts (`staleTime` 60s) — nothing new on any content-script page-load path.

**i18n**

All 9 locales: "(Beta)" stripped from the request label, `betaRequired`/`betaApply` replaced with `upgradeRequired`/`upgrade`, new pool labels (`pools.subscription` / `pools.launchBonus`), `resetsOn` / `expiresOn`, and the three new error strings.

## Related Issue

Closes #

## How Has This Been Tested?

- [x] Added unit tests

Full suite: 281 files / 2674 tests green; oxlint type-aware clean. Beta-specific access-guard cases removed; new cases cover the durationSec payload and each new error branch.

## Checklist

## Additional Information

**⚠️ Draft until `@read-frog/api-contract@0.13.0` publishes.** `package.json` still pins `0.12.6`, whose strict `create` schema would reject `durationSec` client-side and which lacks `pools`/`plan` on the usage type — CI typechecks against the published package and will stay red until the monorepo release lands. Local dev and this PR's verification ran with `WXT_USE_LOCAL_PACKAGES=true` (monorepo source aliases). Unblock steps once 0.13.0 is on npm:

1. Bump `@read-frog/api-contract` to `0.13.0` (+ `@read-frog/definitions` if released together), `pnpm install`
2. Re-run CI, mark ready for review

Ship order: server PR merges + deploys + release publishes → this PR bumps the dep and lands → store release. An old published extension against the new server keeps working (create without `durationSec` is accepted during the grace window; the beta error code no longer occurs).


mengxi-ream:
Second commit (`0ded4df29`): the fixed 5-minute poll deadline no longer renders a still-running transcription as a failure.

- Deadline = `min(20min, 8min + 0.1 × video length)` — wall time barely tracks video length (parallel Azure chunks settle a 20-min video in ~1 min), so the base term absorbs the real variance: download flakiness + the worker retry chain.
- Expiry now shows a **still transcribing — click again in a few minutes** toast instead of the failure overlay; a later click resumes the same cached row (instantly, if it completed meanwhile). A real `status: failed` from the server still renders the failure overlay, unchanged.
- Polling stays at a fixed 1s — a few primary-key reads per second isn't worth a backoff's complexity. If concurrent transcriptions ever sustain >50 or `videoTranscript.get` dominates API traffic, the escape hatch is an additive `waitMs` long-poll on `get` backed by BullMQ `QueueEvents` (discussed, deliberately deferred).
- Also fills three error strings missing from zh-TW and fixes an unquoted-colon YAML break in the new es string.

PR #2120, opened 2026-08-23
feat(subtitles): add a scrolling transcript tab to the sidebar

taiiiyang:
> **AI model(s) used (required):** Claude Opus 5 (1M context)

## Type of Changes

- [x] ✨ New feature (feat)
- [ ] 🐛 Bug fix (fix)
- [ ] ✅ Test related (test)
- [ ] 🔄 CI/CD related (ci)
- [ ] 🧠 AI model related (ai)

## Description

Adds a **transcript** tab to the subtitles sidebar: the whole track, one line at a time with its timestamp, following playback and seeking when a line is clicked.

The tab sits first, so opening the panel lands on the transcript and the summary generates once its own tab is opened. The menu entry is renamed **Learning**, since it now leads to two features and the panel's own tab bar is where you choose between them.

### The timeline is the layout

The left column is a continuous rail rather than a list of labels, because the content genuinely is a sequence in time. The rail does two jobs at once: it indexes the lines, and it fills behind the playhead so you can see how far into the video you are. The line being spoken carries a tinted block, a filled node on the rail, and its timestamp in the brand colour; every other node stays hidden until hover so the rail does not become a row of dots.

Every colour comes from the existing theme tokens — this is injected chrome inside someone else's page, and it has to read as the same surface as the subtitles settings panel.

### Following

The active line is centred while following. Following stops on **intent** — wheel, touch, or a scrolling key — rather than on the `scroll` event, because programmatic scrolling fires that too and the panel would switch itself off the moment it moved. A "back to current" button restores it.

### Notes for the reviewer

**Playback time is now published even when no subtitles session is running.** The scheduler returned early from `timeupdate` whenever it was inactive, so the transcript never moved for anyone who had not turned captions on. The guard moved below the write: time is published either way, cue resolution stays gated, and nothing new reaches the player. A test in `subtitles-scheduler.test.ts` pins all three halves of that.

**Translations arrive through a new atom.** The scheduler kept its translated cues private and published only whichever one covers now — enough for the caption overlay, not for a list. `translatedTrackAtom` is written at the three points where that array changes, and the transcript pairs it onto the source track by `start`, which is already how the coordinator identifies a cue.

**The transcript loads the track itself when none is published.** Publishing it does not put captions on the video: rendering is gated on `subtitlesVisibleAtom`, which only the scheduler sets.

**Panel gating was re-split.** Only "this video has no subtitles" still applies to both tabs. A provider that cannot be prompted now says so inside the summary tab — a transcript needs no model and should not be locked behind one.

**No virtualisation.** A long video can run past a thousand rows; this ships as a plain list and gets a virtualised one only if real playback shows it needs it.

**`videoSummaryAtom` is deleted** — it lost its last reference when the summary moved to the query cache, and being an exported symbol it slipped past lint.

### Screenshots
image

## Related Issue

Closes #

## How Has This Been Tested?

- [x] Added unit tests

`src/utils/subtitles/__tests__/transcript.test.ts` covers pairing translations by `start` (missing, mismatched, and untranslated cues) and locating the active line (boundaries, the gap between cues, before the first and after the last). `subtitles-scheduler.test.ts` gains the inactive-time-publication case.

Full suite: **2807 passed**, 4 skipped. `pnpm lint` clean, `pnpm build` succeeds.

Following was measured on a real player by reading the active row's offset from the viewport centre as playback advanced:

| Playback | Active line | Offset from centre |
|---|---|---|
| 0s | 1 | −298px (nothing above it to scroll) |
| 20s | 6 | **0** |
| 60s | 16 | **0** |
| 100s | 26 | **0** |
| 140s | 36 | **0** |

**Still unverified.** YouTube answers `403` for the caption fetch in a fresh automated profile with no session, so the measurements above ran against seeded rows and these need a real browser: loading an actual track, pausing on manual scroll, "back to current", click-to-seek, and the transcript rendering source-only with captions turned off.

## Checklist

- [ ] If my code was generated by AI, I have proofread and improved it as necessary.


src/entrypoints/subtitles.content/ui/subtitles-sidebar/sections/transcript/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+      onClick={resume}
+      className="pointer-events-auto shadow-floating"
+    >
+      {activeAbove ? <IconArrowUp className="size-3.5" /> : <IconArrowDown className="size-3.5" />}
+      {i18n.t("subtitles.sidebar.transcript.backToCurrent")}
+    </Button>
+  )
+
+  return (
+    <div ref={rootRef} className="relative" {...intentProps}>
```
mengxi-ream, line 92: Handle keyboard scroll intent on the scroll viewport**

When keyboard focus is on the scrollable viewport, pressing `PageUp`, `PageDown`, or an arrow key scrolls the transcript without pausing following. Here `intentProps` is attached to the inner transcript div, but the [scroll viewport is its ancestor](https://github.com/mengxi-ream/read-frog/blob/f66390900b109a351b0eda260ef070809eaa87a9/src/components/ui/base-ui/scroll-area.tsx#L23-L31), so key events targeting the viewport never reach this descendant handler. `following` stays true and the next active-cue change calls `scrollIntoView`, overriding the user's manual position without showing “Back to current.”

Reproduced with the actual `ScrollArea` component: dispatch `PageUp` on `[data-slot="scroll-area-viewport"]`, advance playback into the next cue, and `scrollIntoView` is still called. Please listen for keyboard intent on the viewport or scroll-area root, with appropriate cleanup. This is separate from the existing scrollbar-pointer finding.

mengxi-ream (changes requested):
AI 的一些评论还是需要 address 一下，有些是 valid

PR #2202, opened 2026-09-17
feat(subtitles): translate captions on x.com status videos

taiiiyang:
> **AI model(s) used (required):** Claude Opus 5

## Type of Changes

- [x] ✨ New feature (feat)

## Description

Adds subtitle translation for x.com status videos, and with it a third subtitle-acquisition mechanism that sits alongside the existing `fetchers/ai/` and `fetchers/youtube/`.

### A generic TextTrack mechanism

`src/utils/subtitles/fetchers/text-track/` reads cues straight from the video's own `TextTrack` objects. It takes only `resolveVideo` and `getVideoId` from the platform, so any site whose player exposes native text tracks can reuse it.

The load trick is the part worth reviewing. A caption track the viewer never switched on sits at `mode: "disabled"` with `cues === null`. Setting it to `"hidden"` makes the browser fetch the whole VTT — on the tweet used for verification, 78 cues arrived in one go. That means no HLS manifest parsing, no segment stitching, and no `X-TIMESTAMP-MAP` alignment.

Native captions are suppressed by moving `showing` tracks to `"hidden"` as well. **Never to `"disabled"`** — that stops cue loading and would cut off our own source. x.com flips tracks back on its own, so the hide is re-applied after a short delay. Two optional hooks (`hideNativeSubtitles` / `showNativeSubtitles`) were added to `SubtitlesFetcher`; `UniversalVideoAdapter` delegates to them from the two private methods it already had, so the five existing call sites pick this up with no behaviour change for YouTube.

### Screenshot
image

### x.com platform layer

x.com ships no stable player or controls selectors, so `platforms/xcom/overlay-entry.ts` stamps its own attributes onto the tweet's video container and onto the native controls group (found via `[data-testid="captions"]` and two levels up). Every selector in `getXcomConfig` targets those stamps, which keeps `UniversalVideoAdapter` untouched.

That controls group is **discarded and rebuilt by x.com on every hide** — measured at 11 rebuilds over 10 hover cycles. Two consequences are handled:

- The button host is tracked by reference, because once the group is discarded the host is detached and unreachable from the document. Its React root is unmounted through the existing `removeReactShadowHost` before a new host is created; otherwise each rebuild would leak a root whose `matchMedia` listener and store subscription pin the detached tree (the same failure documented in `page-translation.ts`).
- A `MutationObserver` on the video container (50 ms debounce) re-mounts the button as soon as the group reappears. Observing the container rather than the document keeps it cheap on a timeline that mutates while scrolling. The existing poll stays as a 2 s fallback.

AI subtitles are omitted on x.com rather than shown disabled: the transcription server only accepts YouTube URLs, so the menu entry would never work.

### Scope

Only one video per document is handled. Timeline and media-tab pages mount nothing — `pickSingleCandidate` declines when several videos are visible. Support for those is a follow-up.

## How Has This Been Tested?

- [x] Added unit tests

Unit tests cover the mechanism with duck-typed `TextTrack` and video objects, so they touch no real DOM: track selection, `disabled → hidden` cue loading, waiting for late cues, track identity hashing, re-hide after the player flips a track back, and the status-id extractor (including look-alike hosts and `/analytics` suffix stripping).

Manual verification against live x.com in Chrome 152, driven through the running dev browser:

| Check | Result |
|---|---|
| Cues from a track the viewer never enabled | 78 loaded |
| Button insertion after a controls rebuild | 60 ms median over 10 cycles |
| Button host count after 10 rebuilds | stays at 1 |
| Injected `<style>` accumulation | none |
| Translation | bilingual output rendered on the video |
| Native captions | hidden while ours are on, restored when off |
| AI subtitles entry | absent |
| YouTube `/watch` | no regression |

`pnpm exec tsc --noEmit` clean, `oxlint --type-aware` clean, 331 subtitle tests passing.

## Checklist


taiiiyang:
### Code review

Found 1 issue:

1. The x.com translate button has two owners that do not know about each other, and the tracked reference goes stale, so the real button leaks on cleanup (bug due to `mountedButton ?? document.getElementById(...)` in `overlay-entry.ts` and `renderTranslateButton()` in `universal-adapter.ts`)

`mountXcomTranslateButton()` creates the button and stores it in the module-level `mountedButton`. `UniversalVideoAdapter.renderTranslateButton()` also targets `selectors.controlsBar` (set to the x.com controls stamp), runs on `initialize()` and on every `notifyNavigation()`, finds that button, calls `removeReactShadowHost()` on it and creates a new host, but never updates `mountedButton`. `mountXcomTranslateButton()`'s dedupe check then sees the new button and returns without re-claiming the ref. Later `removeTranslateButton()` reads `mountedButton ?? document.getElementById(...)`; the stale ref is non-null, so `??` never falls back and the currently mounted button is never unmounted. `init-xcom-subtitles.ts` calls `notifyNavigation()` on every video change, so this happens during ordinary timeline scrolling. Simplest fix: give the button one owner. Either let the adapter's generic mount path be the only creator (x.com only stamps the controls bar, the way YouTube works), or have `mountXcomTranslateButton()` re-claim `mountedButton` from the DOM when it finds an existing host.

https://github.com/mengxi-ream/read-frog/blob/fa07f70929dc59b38459a85ed66810f037103232/src/entrypoints/subtitles.content/platforms/xcom/overlay-entry.ts#L36-L46

https://github.com/mengxi-ream/read-frog/blob/fa07f70929dc59b38459a85ed66810f037103232/src/entrypoints/subtitles.content/platforms/xcom/overlay-entry.ts#L74-L83

https://github.com/mengxi-ream/read-frog/blob/fa07f70929dc59b38459a85ed66810f037103232/src/entrypoints/subtitles.content/universal-adapter.ts#L569-L582

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

mengxi-ream (approved):

